PR #9, opened 2019-02-01
Dm 131 - User login/register

RescuePenguin:
https://github.com/devise-security/devise-security for reference


test/models/user_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
```
ahay-agile6: delete this file later

ahay-agile6:
TODO: add user model tests

PR #10, opened 2019-02-08
DM-140 User role implementation

RescuePenguin:
User roles + confirmable authentication

No User role permissions are in this ticket, but I have added a quick "authentication" process using the Confirmable devise plugin.


app/models/user.rb
```diff
@@ -36,4 +40,10 @@ def valid_email
 
     errors.add :email, 'invalid'
   end
+
+  def remove_all_roles
```
ahay-agile6: this callback has one argument sent in to it (`roles`)

db/seeds.rb
```diff
@@ -223,6 +223,11 @@
     ]
   end
 
+  ### USERS ###
+  User.create!(email: 'tom.black@agile6.com', password: 'Password123', password_confirmation: 'Password123').add_role(User::USER_ROLES[1].to_sym)
```
ahay-agile6: if we're going to confirm email addresses, tom's needs to be correct: `tom@skylight.digital`

app/models/user.rb
RescuePenguin: But it doesn't use the argument. So It doesn't need to be declared.

PR #12, opened 2019-02-13
DM-138 && DM-146 Permissions for unauthenticated users and approver/editor and admin for practices

ahay-agile6:


config/initializers/devise.rb
```diff
@@ -18,7 +18,7 @@
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
+  config.mailer_sender = 'diffusionmarketplace_donotreply@va.gov'
```
ahay-agile6: this will need to be updated later

app/models/practice.rb
```diff
@@ -1,10 +1,12 @@
 class Practice < ApplicationRecord
   acts_as_list
   # has_attached_file :main_display_image, styles: { medium: "300x300>", thumb: "100x100>" }, default_url: "/images/:style/missing.png"
-  has_attached_file :main_display_image, default_url: "/images/:style/missing.png"
-  has_attached_file :origin_picture, default_url: "/images/:style/missing.png"
+  has_attached_file :main_display_image
```
ahay-agile6: we're just putting a color for the background. this throws an error since we don't have that placeholder image

app/models/practice.rb
```diff
@@ -1,10 +1,12 @@
...
+  has_attached_file :origin_picture
```
ahay-agile6: we're hiding this if there is no picture

app/models/practice.rb
```diff
@@ -1,10 +1,12 @@
...
   validates_attachment_content_type :main_display_image, content_type: /\Aimage\/.*\z/
 
+  belongs_to :user, optional: true
```
RescuePenguin: Are we worried about this being a `has_and_belongs_to_many`? Just wondering if multiple contributors will be a thing.

ahay-agile6: we will have to ask andy that soon. for now, it belongs to one user

PR #14, opened 2019-02-15
Dm 150

RescuePenguin:


app/views/users/index.html.erb
```diff
@@ -0,0 +1,26 @@
+<div class="dm-background-linear-up-lightblue-to-white full-view-height">
+  <section class="grid-container x3-top">
+    <fieldset class="usa-fieldset">
+      <input type="checkbox" name="show_disabled" id='show_disabled_users' class='usa-checkbox-input' value='show'>
+      <label class='usa-checkbox-label' for='show_disabled_users'>Show disabled users</label>
+    </fieldset>
+    <table name='Users' class="usa-table" style='width: 100%;'>
```
ahay-agile6: what is the `name` attribute for?

ahay-agile6: let's try to not use the `style` attribute

PR #16, opened 2019-03-07
Add flash messages to admin user management

blacktm:
[DM-197](https://agile6.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=DM3&modal=detail&selectedIssue=DM-197)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -12,6 +12,7 @@
   <body>
     <%= render 'shared/header' %>
     <main id="main-content" class="<%= params[:controller] %>">
+      <%= render 'shared/messages' %>
       <%= yield %>
     </main>
     <%= render 'shared/footer' %>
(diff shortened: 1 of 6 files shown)


app/views/layouts/application.html.erb
```diff
@@ -12,6 +12,7 @@
   <body>
     <%= render 'shared/header' %>
     <main id="main-content" class="<%= params[:controller] %>">
+      <%= render 'shared/messages' %>
```
ahay-agile6: hmmm this might make it so the other pages have unexpected behavior, such as the login page when erroring out and such

blacktm: Yea, just noticed that `<%= render "shared/messages" %>` was already the `home/index` view. For admin alerts, we could also render the partial in `users/index`, or just have only one call in `layouts/application` for global alerts. Got a preference?

blacktm:
Pushed an alternative solution. Message rendering removed from `layouts/application`, now exists in both `home/index` and `users/index`. Merged the the `notice` message type with a more general `flash` that iterates through and renders each flash message, choosing the appropriate [alert styling](https://v2.designsystem.digital.gov/components/alerts/).

PR #21, opened 2019-03-20
Update app to match current designs

blacktm:
For [DM-218](https://agile6.atlassian.net/browse/DM-218)


app/views/shared/_header.html.erb
```diff
@@ -78,7 +75,7 @@
           <% else %>
             <li class="usa-nav-secondary-item">
               <%= link_to new_user_registration_path, class: 'dm-font-color-primary-blue' do %>
-                <%= icon('fas', 'user-circle')%> Register
+                <%= icon('fas', 'user-circle')%> Sign up
```
ahay-agile6: hey @blacktm please make sure this is correct: "Register" vs. "Sign up". i remember feedback from users was that they didn't like "Sign up" because it sounded like email spam. that's why we put "Register" on the site instead. this may not have been changed in the designs

blacktm: Good catch. It is still in the designs, so I'll ask and check.

blacktm: Going to drop this commit and rewrite it, leaving in "Register" for now.

PR #28, opened 2019-04-22
Import dummy data

blacktm:
[DM-325](https://agile6.atlassian.net/browse/DM-325)

This expects a CSV file with the path: `/tmp/practices_dummy_data.csv`
Dummy data CSV is attached in the Jira issue linked above.
The headings should be:

| project_name  | start_date  | business_office_sponsor  | purpose  | approach  |
|---|---|---|---|---|
|   |   |   |   |   |

### DIFF
--- a/lib/tasks/dm.rake
+++ b/lib/tasks/dm.rake
@@ -25,4 +25,26 @@ namespace :dm do
     Rake::Task['dm:full_import'].execute
   end
 
+  # rails dm:dummy_data
+  desc 'Import dummy data'
+  task :dummy_data => :environment do
+    puts '==> Importing dummy data'
+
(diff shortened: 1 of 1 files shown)


lib/tasks/dm.rake
```diff
@@ -25,4 +25,28 @@ namespace :dm do
...
+        name: row['project_name'],
+        tagline: row['project_name'],
+        date_initiated: row['start_date'],
+        description: row['purpose'],
+        summary: row['approach'],
+        initiating_facility: row['business_office_sponsor']
+      )
+
+      practice.approved = true
+      practice.published = true
```
ahay-agile6: how come these aren't in the `create` arguments?

blacktm: Ah, no reason, a victim of too much refactoring. Will push an update.

PR #131, opened 2020-03-07
Dm 1426

ahay-agile6:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-1426

users were seeing favorited practices in the "created practices" section.
and if they unfavorited, they still see it in the "created practices" section.
this is because of how the practices are set up.
Users can have many practices through UserPractice, but this is only for "committed" or "favorited" practices.
Instead, we need to look at the Practice's relationship with the user. a Practice can belong_to a user, that's how we know it was created by them

1. favorite a practice
2. make sure it does not show up in the "Create Practices" section of the user profile page if the user is not the owner (`Practice.user`)

Before...
I have Flow3 favorited, and I had Nalaxone favorited and unfavorited, but I still created it? lol

localhost_3200_users_2

After!
I only have Flow3 favorited and I'm not the owner of any of practice, which is correct :)

localhost_3200_users_2 (1)

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -72,14 +72,15 @@
     </section>
   <% end %>
 
-  <% if current_user && current_user.practices.any? %>
+  <% created_practices = Practice.where(user: current_user) %>
+  <% if current_user && created_practices.any? %>
     <section class="dm-background-white margin-bottom-105">
       <div class="grid-container">
         <div class="grid-row">
           <h2 class="profile-h2 font-sans-xl margin-top-0">Created Practices</h2>
         </div>
         <div class="grid-row grid-gap-2 created-practices margin-bottom-5">
-          <% current_user.practices.each do |cup| %>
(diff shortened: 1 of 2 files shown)


app/views/users/show.html.erb
```diff
@@ -72,14 +72,15 @@
     </section>
   <% end %>
 
-  <% if current_user && current_user.practices.any? %>
+  <% created_practices = Practice.where(user: current_user) %>
```
emyl3: Question for learning purposes 😅: Is there a preference for one variables should be set in the controller vs the view? 🤔

ahay-agile6: good question.
I don't have a system for this...
I think typically you're only supposed to have one "instance" variable
(that's the variables prefaced with `@`)
However, setting things in the view fixes that problem.
I don't mind either way as long as code is readable (or obvious) and/or explained or commented out :wink:

PR #184, opened 2020-05-15
Dm 1687

bradjohnson92008:
### JIRA issue link
dm-1687
https://agile6.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=DM&modal=detail&selectedIssue=DM-1687&assignee=5af333511864357363fa79d1

Introduces new scope (filter) to admin/practices page for 'Get Practice Owner Emails'

1.  navigate to /admin/practices
2.  click on the "Get Practice Owner Emails' button
3.  verify the column Headers for the data are "Practice Name" and "Owner Email" and data displays.
4.  click on the "CSV" button at the bottom of the page.
5.  verify data is downloaded to local machine.
6.  verifty data in the csv contains the following headers "Practice name" and "Owner email" and contains same data as displayed on the page.

image

## Acceptance criteria

- Add button in Active Admin called “Get Practice Owner Emails”
- Put button on the top
- Fields in spreadsheet/CSV: Practice Owner Email Address and Practice Name 
- Button creates a list in active admin (optional)
- Button creates an csv that is downloadable
- csv has a row with practice names
- csv has row with practice editors' email addresses 

## Definition of done


app/admin/practices.rb
```diff
@@ -79,8 +175,8 @@ def set_practice_user(practice)
       practice.commontator_thread.subscribe(user)
     end
 
-
     def find_resource
+      debugger
```
ahay-agile6, line 179: remove :)

app/models/practice.rb
```diff
@@ -111,6 +111,7 @@ def origin_picture_s3_presigned_url(style = nil)
 
   scope :published,   -> { where(published: true) }
   scope :unpublished,  -> { where(published: false) }
+  scope :Get_Practice_Owner_Emails, -> {where(user_id: exists?)}
```
ahay-agile6, line 114: the ruby way: let's lower case this

app/admin/practices.rb
```diff
@@ -5,14 +5,110 @@
...
+  end
 
   index do
-    selectable_column
-    id_column
-    column :name
-    column :support_network_email
-    column(:user) {|practice| practice.user&.email}
-    column :created_at
+    Practice.create
```
ahay-agile6, line 105: question: what does this do?

app/admin/practices.rb
```diff
@@ -5,14 +5,110 @@
 
   scope :published
   scope :unpublished
+  scope :Get_Practice_Owner_Emails
```
ahay-agile6, line 8: the ruby way: let's lower case this

app/admin/practices.rb
```diff
@@ -5,14 +5,110 @@
...
+
+  csv do
+    if params[:scope] == "get_practice_owner_emails"
+      column(:practice_name) {|practice| practice.name}
+      column(:owner_email) {|practice| practice.user&.email}
+    else
+      column :id
```
ahay-agile6, line 15: wondering if we can do something fancy here so we don't have to update it when we add columns...

app/models/practice.rb
bradjohnson92008: good catch... I didn't think Ruby was smart enough to upper case the text for the button display.. :)

app/admin/practices.rb
bradjohnson92008: this just renders a practice for each one that is fetched by the current scope/filter or query that is run by active admin... and displayed in the "Practice" view...  based on the criteria below that code.. i.e., which columns to display.... etc.

app/admin/practices.rb
bradjohnson92008: Got it... i've updated both of those instances to lower case and have pushed.

app/admin/practices.rb
bradjohnson92008: Yeah - I was looking for a way to default to all columns in that else case - ie.., columns.* or something - but didn't find one.   I can do some more research on that today.

app/admin/practices.rb
bradjohnson92008: done

app/admin/practices.rb
bradjohnson92008: I figured out how to add the column names dynamically - so for published and unpublished scopes/filters.. all column names in the Practice Table will be included in the csv.  Code has been pushed to server for dm-1687

PR #287, opened 2020-11-18
Dm 2181

bradjohnson92008:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-2181

This code sends an email to dm@agile6.com whenever Covid related categories are selected or unselected for a practice.  Email is sent after clicking the save button.  One email is generated which details all (covid) categories that were either newly selected or unselected.

1. log in to DM as admin.
2.  Select a practice and go to the Practice Editor - Introduction page.
3.  Scroll down to the categories.
4.  Select or unselect covid related categories (see AC below).
5.  Click on Save.
6.  check your local machine DM path/tmp/letter_opener folder - the generated email should be in this folder.
7.  Open the email/file in letter_opener folder that matches the time you clicked on save and view.
8.  Verify the covid related categories selected or unselected are detailed in the email.
9.  Verity a link to the Practice page exists after the categories are listed.
10.  The email subject should read... "The following covid 19 related categories were added or removed from '<Practice-Name>'".

image

N/A

## Acceptance criteria
Generate an email every time someone selects or removes the follow categories in the PE:

COVID
Telehealth
COVID Testing and Screening
Curbside Care
Environmental Services
Pulmonary Care
Healthcare Administration

Email goes to DM@Agile6.com ideally. If not, Marketplace@va.gov
Generate the email when they hit “Save” 
Email details the practice and which category they selected
Email contains a link to the practice
If a user removes a category, same workflow as add (except email says “Removed”

## Definition of done


JoshingYou1 (changes requested):
Nice work Brad!! Everything works as intended for me. I have a couple of suggestions code-wise though. In the `covid_category_mailer.rb` file, we should use interpolation instead of concatenation. I also thought we could added in a bit of dynamic text to the subject of the email, but adding that part is up to you. Here's what I had in mind: 

```
# Mailer for form emails.
class CovidCategoryMailer < ApplicationMailer
  default from: ENV['DO_NOT_REPLY_MAILER_SENDER'] || 'do_not_reply@va.gov'
  layout 'mailer'

  def send_covid_category_selections(options = {})
    practice_name = options[:practice_name]
    selected_categories = options[:selected_categories]
    url = options[:url]
    text = ""
    if selected_categories.present?
      text = "Selected Categories:\n"
      selected_categories.each do |sc|
        text = "#{text}#{sc.name}\n"
      end
      text = "#{text}\n\n\n"
    end
    unselected_categories = options[:unselected_categories]
    if unselected_categories.present?
      text = "#{text}Unselected Categories:\n"
      unselected_categories.each do |uc|
        text = "#{text}#{uc.name}\n"
      end
    end
    text = "#{text}#{unselected_categories.present? ? "\n\n\n" : ''}Link to #{practice_name} : #{url}"
    added = selected_categories.present? ? 'added' : nil
    removed = unselected_categories.present? ? 'removed' : nil

    subject = "The following covid-19 related categor#{selected_categories.count + unselected_categories.count == 1 ? 'y' : 'ies'}
               w#{selected_categories.count + unselected_categories.count == 1 ? 'as' : 'ere'}
               #{added} #{added && removed ? 'or' : ''} #{removed} #{added && removed || removed ? 'from' : 'to'}
              '#{practice_name}'"
    mail(to: 'dm@agile6.com', subject: subject, body: text)
  end
end
```

Try it out and let me know what you think. Other than that, awesome job on this! :grin:

app/services/save_practice_service.rb
```diff
@@ -145,6 +145,55 @@ def update_category_practices
     category_params = @practice_params[:category]
     practice_category_practices = @practice.category_practices
     practice_categories = @practice.categories
+    categories_selected = Array.new
```
emyl3, line 148: can we move this mailer logic outside of this `update_categories_practices` method to a separate one that just handles this?

app/services/save_practice_service.rb
```diff
@@ -145,6 +145,55 @@ def update_category_practices
...
+    categories_unselected = Array.new
+    dm_notification_categories = ["COVID", "TELEHEALTH", "CURBSIDE CARE", "ENVIRONMENTAL SERVICES", "PULMONARY CARE", "HEALTHCARE ADMINISTRATION"]
+
+
+    if category_params.blank? && practice_categories.present?
+      practice_categories.each do |pc|
+        categories_unselected.push(pc.id)
```
emyl3, line 155: Could we consolidate some of the logic here so instead of pushing it to one array and then iterating over that array again?
- We could get an array of the `dm_notification_categories` by finding `Category` with that `name` then pushing that `Category` `id` into that array.
- Then we can check if the id is a covid category before pushing it in.
Let me know  what you think, Brad!

emyl3 (commented):
Let some ideas for code refactoring. Let me know what you think, Brad! Otherwise worked as intended. :)

app/services/save_practice_service.rb
bradjohnson92008: Yeah - thats a good call Elisa - I'll move it all in to it's own method.

bradjohnson92008: Yeah thats cleaner - being in a separate method. thx

app/services/save_practice_service.rb
bradjohnson92008: I see what you mean Elisa... use the category IDs instead of name but I'm doing a contains on "COVID" to catch any future covid related categories...  plus there are only 6 covid related categories so it's not a performance issue.. I agree if we were comparing 1000s of strings in a list.. then using array of ids would be much faster but you'd still have the upfront work of creating the array of category IDs..

bradjohnson92008:
Josh - I made the changes you suggested regarding interpolation vs concatenation.

JoshingYou1 (approved):
Works great! Nice job on this Brad!

emyl3 (approved):
Thank you, Brad!

PR #296, opened 2020-12-03
Dm 2236

bradjohnson92008:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-2236

1.  Adds the date_published field to the practices table
2.  Updates the date_published field with current timestamp if practice is successfully published.
3.  Adds the "Date Published" column to the /admin/practices view with DESC and ASC sort capability.

1.  Login in to DM as an admin.
2.  navigate to /admin
3.  select the practices tab at the top of the page.
4.  Verify that the "Date Published" field exists.
Note* Since you haven't published any practices since downloading this branch all the values Date Published will default to null.
5.  Go to the practice Editor for a practice in your DB that has not been published.
6.  Fill in all required fields for publishing.
7.  Click on the "Publish" button. (if validations do not succeed - go back to step 6 :)
8.  Verify that the Practice was published (should have confirmation banner at top of PE page).
9.  Navigate back to /admin path and click on Practice Tab.
10.  The Date Published field should be populated for the practice you just published.
11.  Click on the "Date Published" column header.  The default sort order is descending.
12.  Verify that the Practice you just published is listed first.  (Null values are displayed last).
13.  If you click the Date Published header again the practices will be sorted in Date Published ASC order so the NULLs will display first and the row for the practice you just published will be last.

image

N/A

## Acceptance criteria
- Add Date Published to DB
- Add it on "Publish" 
- Make it a sortable category in Admin panel
- Make it usable for "Recently Published" should we add that as a configured experience 

## Definition of done
- [X ] Unit tests written (if applicable)
- [X ] A link has been provided to the originating JIRA issue
- [X ] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

### DIFF
--- a/app/admin/practices.rb
+++ b/app/admin/practices.rb
@@ -41,7 +41,7 @@ ActiveAdmin.register Practice do
     column :enabled unless params[:scope] == "get_practice_owner_emails"
     column :created_at unless params[:scope] == "get_practice_owner_emails"
     column 'Last Updated', :updated_at unless params[:scope] == "get_practice_owner_emails"
-    column 'Date Published', :date_published unless params[:scope] == "get_practice_owner_emails"
+    column :date_published unless params[:scope] == "get_practice_owner_emails"
     actions do |practice|
       practice_enabled_action_str = practice.enabled ? "Disable" : "Enable"
       item practice_enabled_action_str, enable_practice_admin_practice_path(practice), method: :post
--- a/app/admin/practices.rb
+++ b/app/admin/practices.rb
@@ -24,6 +24,14 @@ ActiveAdmin.register Practice do
       ['lower(practices.name)', order_clause.order].join(' ')
     end
 
+  order_by(:date_published) do |order_clause|
+    if order_clause.order == 'desc'
+      [order_clause.to_sql, 'NULLS LAST'].join(' ')
+    else
+      [order_clause.to_sql, 'NULLS FIRST'].join(' ')
+    end
+  end
+
     index do
     selectable_column unless params[:scope] == "get_practice_owner_emails"
     id_column unless params[:scope] == "get_practice_owner_emails"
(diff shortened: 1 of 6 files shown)


app/admin/practices.rb
```diff
@@ -33,6 +41,7 @@
     column :enabled unless params[:scope] == "get_practice_owner_emails"
     column :created_at unless params[:scope] == "get_practice_owner_emails"
     column 'Last Updated', :updated_at unless params[:scope] == "get_practice_owner_emails"
+    column 'Date Published', :date_published unless params[:scope] == "get_practice_owner_emails"
```
emyl3, line 44: instead of 'Date Published' you can just do `column :date_published unless params[:scope] == "get_practice_owner_emails"`

app/services/save_practice_service.rb
```diff
@@ -24,6 +24,10 @@ def initialize(params)
 
   def save_practice
     begin
+      if @practice_params["support_network_email"].present? && @practice.approved == false
```
emyl3, line 27: I am curious as to why this is necessary 🤔

emyl3 (commented):
Works for me as you described Brad! I just left a couple of comments! :)

app/admin/practices.rb
bradjohnson92008: Yeah - I didn't think of that... but I know.. Active Admin would default to what I wanted in that case..  :)

app/services/save_practice_service.rb
bradjohnson92008: So I had practices that were created a while back without a user email... when creating those in Active Admin if there is no user email the practice would not be set to approved.  We've since made a change - so that even if you don't provide a user email in the initial creation the practice gets set to approved.  Since I had old practices that weren't approved initially - whenever I went to publish them.. (even though I had all the required fields set).. I couldn't approve.  TBH that approved flag may be obsolete.. but I thought that there could be practices in the same state on Prod.. i..,e originally created with no user email prior to the change and are not approved - so they would never be able to publish those practices without this code.  Lets talk about this one more... bc like I said the approved flag may be obsolete with the change made to approve practice creation initially.

emyl3: ok I'm going to see how many practices on prod are not approved

emyl3: there is only 1 practice on prod that is not approved and it isn't published currently... so I think this code can be removed and if there are any issues like this we can do a one-time data fixup. What are your thoughts on this, brad? 🤔

emyl3 (changes requested):

app/services/save_practice_service.rb
bradjohnson92008: Thats fine I can remove that check - If they try to publish that practice - we may just have to go in and manually approve it.   Going forward no practices should fall in to this state.

emyl3 (approved):
Awesome! thank you brad!

JoshingYou1 (approved):
Nice work Brad!! Works for me!!

PR #307, opened 2021-01-19
Dm 2286

bradjohnson92008:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-2286

This code replaces a resource ("File" i.e., media_type=1) name and/or description that is blank.  When we first released resources in the PE implementation we did not require name and description fields to be populated.  This has been implemented now - but some users may have blank name and description - so viewing the resource in the show view - does not display anything for the resource.. just a bullet point... i.e.,
image

Note ** if you do not have any records in your practice_resources table with 'media_type = 1' then go to the PE-> implementation section for a practice and create a File resource.  FYI - name and description are now required so you will have to add data to those fields in order to submit.
image

1.  Update the name and description (set to blank "") for a practice_resources record with "media_type = 1".   Note the practice ID for that record.
2. Go to the Show view for that practice and scroll down to the implementation section.
3. Verify that the resource whose name and description are blank, display the file_name_attachment so that the user can click the resource link and view the file resource.
image

## Acceptance criteria
Fix the blank file and default to attachment file name for name and description if they do not exist on show page

## Definition of done


app/views/practices/show/implementation/_resource_file_attachment.html.erb
```diff
@@ -3,8 +3,12 @@
   <% file_attachments.each do |fa| %>
     <div class="margin-bottom-1">
       <% if fa.attachment.present? %>
+        <%
+          fa.name.length == 0 ? s_name = fa.attachment_file_name : s_name = fa.name
```
emyl3, line 7: if we do `.length` on something that doesn't exist we are going to error out... so maybe we can do 
```
s_name = fa.name ? fa.name : fa.attachment_file_name
s_description = fa.description ? fa.description : fa.attachment_file_name
```

emyl3 (commented):
noticed one thing brad! 😄

app/views/practices/show/implementation/_resource_file_attachment.html.erb
emyl3: we can shorten even more to
```
 s_name = fa.name || fa.attachment_file_name
s_description = fa.description || fa.attachment_file_name
```
i just realized hahah

bradjohnson92008: Hey Elisa  I see what you are trying to prevent - but neither of those work.. bc the name and description fields are non - nullable string fields (char varying)...  which doesn't require a length... so in both cases above... s_name would be set to fa.name which could be an empty string (not null).  i.e.,
image

emyl3: oo gotcha then could we do...
```
s_name = fa.name.blank? ? fa.attachment_file_name : fa.name
s_description = fa.description.blank? ? fa.attachment_file_name : fa.description
```

bradjohnson92008: Yeah - that works as well.  pushed those changes.

JoshingYou1 (commented):
Nice work on this Brad!! The only thing I can recommend is that we write a unit test for this use case. Other than that, everything worked for me as intended!! :smiley:

emyl3:
> Nice work on this Brad!! The only thing I can recommend is that we write a unit test for this use case. Other than that, everything worked for me as intended!! 😃

Good call, Josh!

bradjohnson92008:
Added a unit test for this case.

JoshingYou1 (approved):
Looks good to me. Awesome job Brad!!

PR #318, opened 2021-02-25
DM-2324: Add created practices to profile & recommend for you pg

emyl3:
### JIRA issue link
[DM-2324](https://agile6.atlassian.net/browse/DM-2324)

This PR adds a user's practices that they are able to edit to their Profile page and Recommended for You page.

1. Log in as a user with practices that you can edit (It should be an entry on the PracticeEditor table and not because of your role like an admin)
2. Ensure all those practices appear on your Profile page and Recommended for You page.
3. Remove yourself as an editor on one of those practices.
4. Check that the practice has been removed from the Profile page and Recommend for You page.

## Acceptance criteria
- [X] When Editor is added, the practice is added to “My Practices” on the user’s Profile Page
- [X] Add “Practices Created” to Recommended for you page
- [X] Same info populates on profile and Recommended for you pages under “Practices Created”

## Definition of done

### DIFF
--- a/app/models/practice_editor.rb
+++ b/app/models/practice_editor.rb
@@ -25,4 +25,13 @@ class PracticeEditor < ApplicationRecord
       throw :abort
     end
   end
+
+  def self.get_editor_practices(user)
+    editor_list = PracticeEditor.where(user: user)
+    practices = []
+    editor_list.each do |e|
+      practices.push(Practice.find(e.practice_id))
+    end
+    practices
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -74,6 +74,15 @@ class User < ApplicationRecord
     "#{email}"
   end
 
+  def editor_practices
+    editor_list = PracticeEditor.where(user_id: id)
+    practices = []
+    editor_list.each do |e|
+      practices.push(Practice.find(e.practice_id))
+    end
+    practices
+  end
+
   def password_uniqueness
     return true unless encrypted_password_changed?
     return true if password.split('').uniq.length > 6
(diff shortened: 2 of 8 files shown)


app/models/user.rb
```diff
@@ -74,6 +74,15 @@ def to_s
     "#{email}"
   end
 
+  def editor_practices
+    editor_list = PracticeEditor.where(user_id: id)
```
emyl3, line 78: ⚠️ Please let me know if there is a better way to get all the user's practices via the PracticeEditor relationship 🤔

emyl3 (commented):
⚠️ Now that I think about it... I am wondering if we should include practices that a user is an owner of but not necessarily can edit since it is called "Created practices"... But we can also stick to the AC for now and if it is confusing for users we can change it then. Let me know what you think.

app/models/user.rb
JoshingYou1: Here's a simpler way. Not sure about better, but try it out and let me know what ya think!

```
PracticeEditor.where(user: self).collect { |pe| pe.practice }
```

app/models/practice_editor.rb
```diff
@@ -25,4 +25,13 @@ def ensure_practice_has_at_least_one_practice_editor
...
+
+  def self.get_editor_practices(user)
+    editor_list = PracticeEditor.where(user: user)
+    practices = []
+    editor_list.each do |e|
+      practices.push(Practice.find(e.practice_id))
+    end
+    practices
+  end
+end
```
JoshingYou1, line 37: Are we using this? I didn't see any references to it, that's why I'm asking.

JoshingYou1 (commented):
Everything looks great Elisa! Just left a few comments. Awesome work!!! :smiley:

app/models/practice_editor.rb
emyl3: haha nope! Nice catch I just deleted it now :)

app/models/user.rb
emyl3: ooo nice TIL! thank you!

JoshingYou1 (approved):
LGTM. Great work!

PR #337, opened 2021-04-21
DM-2356: Add created practices search to facility show

emyl3:
### JIRA issue link
[DM-2356](https://agile6.atlassian.net/browse/DM-2356)

This PR adds the created practices search for a facility page.

**Prereq**
1. Set up a facility so that it has 3 or fewer practices that were created at that facility
2. Ensure those practices have some adoptions

**How to Test**
1. Go to the originating facility page of the practices you modified above
2. Ensure the defaults show the following
- 3 results count
- Default to A to Z sort
- No search query
- No categories selected
- No load more button
3. Ensure the sorting works as expected
4. Search for a category name or related term and ensure you see the expected practices and the result count is correct
5. Search for a term in the practices' name, tagline, summary, overview problem, overview solution, overview results, diffusion status (e.g. emerging, replicating, etc), category name, category related term, facility name where practice was adopted, other facility names of originating location
6. Ensure you can filter by categories
7. Ensure when the results are loading there is a loading spinner
8. Ensure you see a no results message when there are no results
9. Repeat 3-8 after adding more practices to the facility so that you see a load more button

[Figma](https://www.figma.com/file/fNXjZzLUS39bKmjdfDWyxy/My-VISN%2FVAMC?node-id=2403%3A4050)

## Acceptance criteria
- [X] VAMC has search for practices created, keyword, and category
- [X] 3 to start then load more
- [X] Sort dropdown (most adoptions, most recently created, and A to Z) 

## Definition of done


JoshingYou1 (commented):
Fantastic work on this Elisa! I only have a couple of small comments.

**Nit 1:**  Ellipses are not being applied to the origin facilities at the bottom of the practice cards.
Screen Shot 2021-04-22 at 1 31 15 PM

**Nit 2:** We should remove the sort select option if a search only returns one result.
Screen Shot 2021-04-22 at 1 31 27 PM

bradjohnson92008:
Great job Elisa - this is looking good.  Just a couple things..
1.  I assume the sort order A-Z is based on the practice's name property.. that doesn't seem to be working. 
image
2.  Where we display the number of results.. wondering if we should display some thing like.. "Displaying 3 of 7 results" instead of "Displaying 7 results" when we're only displaying the first 3...
image
3.  Should the Search button click... event be triggered by default if user enters in search term then hits "Enter".?

bradjohnson92008 (changes requested):
Great Job Elisa.  See my questions above.  BTW... regarding the sort order... 'a-z' .. I think it's working... I checked and the practice that is being listed first (Brads 9th practice" ... has a " " space char in front of it... :)  So up to you if you want to trim the white space or not... probably a very edge case.

emyl3:
> Great job Elisa - this is looking good. Just a couple things..
> 
>     1. I assume the sort order A-Z is based on the practice's name property.. that doesn't seem to be working.
>        image
> 
>     2. Where we display the number of results.. wondering if we should display some thing like.. "Displaying 3 of 7 results" instead of "Displaying 7 results" when we're only displaying the first 3...
>        image
> 
>     3. Should the Search button click... event be triggered by default if user enters in search term then hits "Enter".?

For 1 I think we implemented a server-side fix for that so we can no longer create practices with a whitespace in the beginning which is how I think we should handle this.

For 2, I don't think we do that for any of the practice cards on other pages with load more so I'll leave that alone

For 3, Good point! I'll fix that up! Thank you for catching that Brad!

JoshingYou1 (approved):
Everything looks good to me. Awesome work Elisa!!!

bradjohnson92008 (approved):
Looks great Elisa - great job.

PR #357, opened 2021-06-11
DM-2651: Wrote some unit tests and refactored some code

JoshingYou1:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-2651

1. Adds the practice owner to the list of email recipients when a comment is created for a practice, unless they are already subscribed to the thread.
2. Adds the practice's `support_network_email` to the list of email recipients when a comment is created for a practice, unless a `user` exists with the same email address AND that user is already subscribed to the thread.

_**Subscribing/Unsubscribing Practice Owner**_
1. Log in as an admin and visit `/admin/practices/new`.
2. Create a new practice with a user that has a `va.gov` email address (ex: test@va.gov).
3. Open up the rails console.
4. Run `Practice.last.commontator_thread.subscribers` and make sure the only subscriber is the `practice.user`.
5. Update the same practice with a new `va.gov` email user that you can actually log in with.
6. Run the same command and make sure the first `practice.user` is no longer a subscriber and the new `practice.user` is now a subscriber.
7. Log out as the admin, log back in with the user you added in step `5`, and visit the practice's show page and add a comment.
8. Now log out again, log back in as an admin, and visit /admin/practices/whatever-your-new-practice-slug-is/edit`.
9. Change the `practice.user` to the same user you used in step `2`, and update the practice.
10. Run `Practice.last.commontator_thread.subscribers` and make both `va.gov` users are now subscribers (because the user from step `5` created a comment, even if they are no longer the `practice.user` they should remain subscribed to the thread).

_**Comment Mailer**_
1. Log in as a valid user and visit the show page of a practice of your choice. Make sure the user is NOT the `practice.user` AND that their email does not match the practice's `support_network_email`.
2. Create either a comment or a reply.
3. **If the practice user's email and `support_network_email` are not the same**, make sure the `practice.user`'s email and the `support_network_email` are cc'd on the created comment email.
4. **If the practice user's email and `support_network_email` are the same**, make sure ONLY the `practice.user`'s email is cc'd on the created comment email.
5. Under no circumstance should the comment creator's email be included in the cc list.

## Acceptance criteria
- Practice owner gets email whenever someone comments
- Practice owners gets email every time someone replies to a comment
- Email goes to "Contact" and “Owner” emails

## Definition of done
- [X] Unit tests written (if applicable)
- [X] Events are logged appropriately

### DIFF
--- a/app/models/commontator/subscription.rb
+++ b/app/models/commontator/subscription.rb
@@ -7,7 +7,21 @@ module Commontator
     validates_uniqueness_of :thread_id, :scope => [:subscriber_type, :subscriber_id]
 
     def self.comment_created(comment)
-      recipients = comment.thread.subscribers.reject{|s| s == comment.creator}
+      practice = Practice.find(comment.thread.commontable_id)
+      support_network_email = practice.support_network_email || nil
+      support_network_email_user = support_network_email.present? ? User.find_by(email: support_network_email.downcase) : nil
+      subscribers = comment.thread.subscribers
+
+      if support_network_email.present? && support_network_email_user.nil?
+        subscribers.push(practice.user, support_network_email.downcase)
+      elsif support_network_email.present? && support_network_email_user.present?
+        subscribers.push(practice.user, support_network_email_user)
+      else
+        subscribers.push(practice.user)
+      end
+      # filter out the comment creator and/or any duplicate subscribers
+      recipients = subscribers.reject { |s| s == comment.creator }.compact.uniq
+
       return if recipients.empty?
 
       mail = SubscriptionsMailer.comment_created(comment, recipients)
--- a/app/admin/practices.rb
+++ b/app/admin/practices.rb
@@ -20,7 +20,7 @@ ActiveAdmin.register Practice do
       end
     end
   end
-  
+
     # ensure lowercase practice names are ordered correctly
     order_by(:name) do |order_clause|
       ['lower(practices.name)', order_clause.order].join(' ')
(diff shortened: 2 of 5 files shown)


bradjohnson92008 (approved):
Everything worked as intended for me!  Great job Josh.

app/models/commontator/subscription.rb
```diff
@@ -7,7 +7,21 @@ class Subscription < ActiveRecord::Base
     validates_uniqueness_of :thread_id, :scope => [:subscriber_type, :subscriber_id]
 
     def self.comment_created(comment)
-      recipients = comment.thread.subscribers.reject{|s| s == comment.creator}
+      practice = Practice.find(comment.thread.commontable_id)
+      support_network_email = practice.support_network_email || nil
+      support_network_email_user = support_network_email.present? ? User.find_by(email: support_network_email.downcase) : nil
+      subscribers = comment.thread.subscribers
```
emyl3, line 13: Just so I am understanding the code correctly. `subscribers` can contain both `User` objects or an `email` as a `String`?

emyl3 (approved):
Left one question for you about the code haha! Great work Josh!

app/models/commontator/subscription.rb
JoshingYou1: Under normal circumstances, yes, `subscribers` are `User` objects. However, in the case in which a practice's `support_network_email` is not an existing `User` in the DB, I'm just adding the email to the array in order to include it in the `@recipient_emails` array in the `setup_variables` method from `subscriptions_mailer.rb`. Due to `commontator_thread.subscribers` not being an `ActiveRecord` collection, the changes do not persist after the request. You can check this in your rails console by getting the `commontator_thread.subscribers` both before and after the request, and they should be identical. I hope this was helpful. If not, we can meet to go over this :smiley:

emyl3: Thank you for the detailed explanation, Josh! Very helpful! :D

PR #376, opened 2021-08-05
DM-DEVOPS: run containers as non-root user

ahay-agile6:
This sets the users for the containers as a non-root user, `nginx`.


mutsuoara (approved):
lgtm

PR #380, opened 2021-08-13
DM-start_appcontainer: docker container start up updates

ahay-agile6:
- ensure nginx is the owner of all public assets in the proxy container
- run all non-startup commands with docker instead of docker-compose


mutsuoara (approved):
lgtm tested on dev server.

PR #388, opened 2021-09-07
dm-2820: Update Facilities

bradjohnson92008:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-2820

This code updates the local va_facilities table with the latest and greatest facility information provided from the VAST data set.  This data set also includes VA Vet Center facilities which will be added to the table but marked as hidden for now.   Facilities provided in the VAST data set are either created if they don't exist or updated.  Not deleting any facilities at this time.

1.  From your rails terminal run the task => rails va_facilities:create_or_update_va_facilities.
2.  Code will output progress info on the facilities as they are being processed - either updated or created.
3.  When finished you should see the following: "All VA facilities have been created or updated in the DB!"
4. Prior to running the task the number of records in the DB should be 1211.  After running the task the total number of facilities should be 1595 - 300 of which are "Vet Center" facilities and their hidden field should be marked as true for all 300.  You can run the following query in postgres to confirm.  select count(*) from va_facilities where official_station_name like '%Vet Center%'
4. Go to the facilities page /facilities and ensure that no facilities with a name that includes "Vet Center' are displayed in the drop down of all facilities.

image

N/A

## Acceptance criteria

- Convert .xls to JSON
- Replace dataset we have which I believe is stored as a JSON with the following file that is update to date: https://docs.google.com/spreadsheets/d/1nQLTxTMyC-cYCmsWSuMiX1CBdvZeCHKrw3Ml6lq7kl4/edit#gid=1702691265
- Make sure all the fields map and we have everything we need, including complexity, rurality, and common name.
- Import but don’t display Vet Centers

## Definition of done
- [ X] A link has been provided to the originating JIRA issue

### DIFF
--- a/app/helpers/visns_helper.rb
+++ b/app/helpers/visns_helper.rb
@@ -3,7 +3,7 @@ module VisnsHelper
 
   def get_facility_locations_by_visn(visn)
     sorted_facility_locations = VaFacility.get_by_visn(visn).get_locations
-
+    sorted_facility_locations = sorted_facility_locations.sort
     location_list = ''
 
     # Add other US territories to us_states helper method array
--- a/app/helpers/visns_helper.rb
+++ b/app/helpers/visns_helper.rb
@@ -2,10 +2,9 @@ module VisnsHelper
   include StatesHelper
 
   def get_facility_locations_by_visn(visn)
-    sorted_facility_locations = VaFacility.get_by_visn(visn).get_locations
-    sorted_facility_locations = sorted_facility_locations.sort
+    sorted_facility_locations = VaFacility.get_by_visn(visn).get_locations.sort
     location_list = ''
-
+    
     # Add other US territories to us_states helper method array
     va_facility_locations = us_states.concat(
       [
--- a/app/models/va_facility.rb
+++ b/app/models/va_facility.rb
@@ -44,7 +44,7 @@ class VaFacility < ApplicationRecord
 
   def self.cached_va_facilities
     Rails.cache.fetch('va_facilities') do
-      VaFacility.all
+      VaFacility.where(hidden: false).order('official_station_name')
     end
   end
 end
--- a/app/models/va_facility.rb
+++ b/app/models/va_facility.rb
@@ -8,8 +8,8 @@ class VaFacility < ApplicationRecord
 
   attr_accessor :reset_cached_va_facilities
 
-  scope :get_by_visn, -> (visn) { cached_va_facilities.where(visn: visn) }
-  scope :get_classification_counts, -> (facility_type) { where(classification: facility_type).size }
+  scope :get_by_visn, -> (visn) { cached_va_facilities.where(visn: visn, hidden: false) }
+  scope :get_classification_counts, -> (facility_type) { where(classification: facility_type, hidden: false).size }
   scope :get_classifications, -> { pluck(:classification).uniq }
   scope :get_ids, -> { pluck(:id) }
   scope :get_locations, -> { order(:street_address_state).pluck(:street_address_state).uniq }
--- a/app/models/va_facility.rb
+++ b/app/models/va_facility.rb
@@ -8,7 +8,7 @@ class VaFacility < ApplicationRecord
 
   attr_accessor :reset_cached_va_facilities
 
-  scope :get_by_visn, -> (visn) { cached_va_facilities.where(visn: visn, hidden: false) }
+  scope :get_by_visn, -> (visn) { cached_va_facilities.where(visn: visn) }
   scope :get_classification_counts, -> (facility_type) { where(classification: facility_type, hidden: false).size }
   scope :get_classifications, -> { pluck(:classification).uniq }
   scope :get_ids, -> { pluck(:id) }
--- a/app/views/va_facilities/_index_table_row.html.erb
+++ b/app/views/va_facilities/_index_table_row.html.erb
@@ -3,7 +3,11 @@
     <th scope="row" role="rowheader" class="grid-col-4"><a class="usa-link" href='<%= va_facility_path(facility) %>'><%= facility["official_station_name"] %> (<%= facility["common_name"] %>)</a></th>
     <td class="grid-col-1"><%= facility.street_address_state %></td>
     <td class="grid-col-1"><a class="usa-link" href='/visns/<%= facility.visn.number %>'><%= facility.visn.number %></a></td>
-    <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level[0..1].upcase %></td>
+    <% if facility.fy17_parent_station_complexity_level.match(/^1/) %>
+      <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level[0..1].upcase %></td>
+    <% else %>
+      <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level[0..0].upcase %></td>
+    <% end %>
     <td class="grid-col-2"><%= facility.practices_created_count %></td>
     <td class="grid-col-2"><%= facility.practices_adopted_count  %></td>
   </tr>
--- a/app/views/va_facilities/_index_table_row.html.erb
+++ b/app/views/va_facilities/_index_table_row.html.erb
@@ -3,11 +3,7 @@
     <th scope="row" role="rowheader" class="grid-col-4"><a class="usa-link" href='<%= va_facility_path(facility) %>'><%= facility["official_station_name"] %> (<%= facility["common_name"] %>)</a></th>
     <td class="grid-col-1"><%= facility.street_address_state %></td>
     <td class="grid-col-1"><a class="usa-link" href='/visns/<%= facility.visn.number %>'><%= facility.visn.number %></a></td>
-    <% if facility.fy17_parent_station_complexity_level.match(/^1/) %>
-      <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level[0..1].upcase %></td>
-    <% else %>
-      <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level[0..0].upcase %></td>
-    <% end %>
+    <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level.match(/^1/) ? facility.fy17_parent_station_complexity_level[0..1].upcase : facility.fy17_parent_station_complexity_level[0..0].upcase %></td>
     <td class="grid-col-2"><%= facility.practices_created_count %></td>
     <td class="grid-col-2"><%= facility.practices_adopted_count  %></td>
   </tr>
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_08_17_165620) do
+ActiveRecord::Schema.define(version: 2021_09_03_160740) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
@@ -964,7 +964,6 @@ ActiveRecord::Schema.define(version: 2021_08_17_165620) do
     t.integer "maturity_level"
     t.datetime "date_published"
     t.datetime "practice_pages_updated"
-    t.string "highlight_title"
     t.string "highlight_body"
     t.boolean "retired", default: false, null: false
     t.string "retired_reason"
@@ -1225,6 +1224,7 @@ ActiveRecord::Schema.define(version: 2021_08_17_165620) do
     t.string "slug"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.boolean "hidden", default: false, null: false
     t.index ["station_number"], name: "index_va_facilities_on_station_number", unique: true
     t.index ["visn_id"], name: "index_va_facilities_on_visn_id"
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -964,6 +964,7 @@ ActiveRecord::Schema.define(version: 2021_09_03_160740) do
     t.integer "maturity_level"
     t.datetime "date_published"
     t.datetime "practice_pages_updated"
+    t.string "highlight_title"
     t.string "highlight_body"
     t.boolean "retired", default: false, null: false
     t.string "retired_reason"
--- a/lib/tasks/va_facilities.rake
+++ b/lib/tasks/va_facilities.rake
@@ -81,7 +81,6 @@ namespace :va_facilities do
             # update record.....
             facility = VaFacility.where(station_number: vaf["Station_Number"]).first
             if facility.present? && visn.number === vaf["VISN"].to_i
-              puts 'facility not blank and visn # good'
               facility.visn = visn
               facility.sta3n = vaf["STA3N"]
               facility.station_number = vaf["Station_Number"]
--- a/lib/tasks/va_facilities.rake
+++ b/lib/tasks/va_facilities.rake
@@ -74,24 +74,25 @@ namespace :va_facilities do
 
 
   task :update_va_facilities => :environment do
-    debugger
     ctr = 0
     file = File.read("#{Rails.root}/lib/assets/va_facilities.json")
-    if valid_json?(file)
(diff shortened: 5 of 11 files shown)


db/schema.rb
```diff
@@ -964,7 +964,6 @@
     t.integer "maturity_level"
     t.datetime "date_published"
     t.datetime "practice_pages_updated"
-    t.string "highlight_title"
```
JoshingYou1, line 967: I don't think we want to remove this until the feature branch is in master

lib/tasks/va_facilities.rake
```diff
@@ -2,60 +2,60 @@ namespace :va_facilities do
...
+            parent_station_number: vaf["Parent_Station_Number"],
+            official_parent_station_name: vaf["Parent_Station_Name"],
+            fy17_parent_station_complexity_level: vaf["Parent_Station_Complexity_Level_2017"],
+            operational_status: vaf["Operational_Status_Active_A_Or_Planned_P_Or_Temporarily_Deactivated_T_Or_Permanently_Deactivated_D"],
+            ownership_type: vaf["Ownership_Type"],
+            delivery_mechanism: vaf["Delivery_Mechanism"],
+            staffing_type: vaf["Staffing_Type"],
+            #va_secretary_10n_approved_date: vaf["VASecretary10NApprovedDate"],
+            #planned_activation_date: vaf["PlannedActivationDate"],
+            #station_number_suffix_reservation_effective_date: vaf["Station_NumberSuffixReservationEffectiveDate"],
```
JoshingYou1, line 29: If we don't need these, we can remove them.

lib/tasks/va_facilities.rake
```diff
@@ -68,7 +68,228 @@ namespace :va_facilities do
...
+    ctr = 0
+    file = File.read("#{Rails.root}/lib/assets/va_facilities.json")
+    va_facilities = JSON.parse(file)
+    visns = Visn.all
+    visns.each do |visn|
+      puts 'VISN: ' +  visn.number.to_s
+      va_facilities.each do |vaf|
+          if visn.number === vaf["VISN"].to_i && VaFacility.where(station_number: vaf["Station_Number"]).empty?
+            puts 'Creating facility - ' + vaf["Official_Station_Name"]
+            hidden = false
```
JoshingYou1, line 85: Do we need this, since we set the default to `false`?

lib/tasks/va_facilities.rake
```diff
@@ -68,7 +68,228 @@ namespace :va_facilities do
...
+            if (vaf["Official_Station_Name"].include? "Vet Center")
+              hidden = true
+              create_vet_center_facility visn, vaf, hidden
```
JoshingYou1, line 88: Can we use parentheses for the parameters here? Also, can we just write `true` for the value of hidden as opposed to writing `hidden = true`?

lib/tasks/va_facilities.rake
```diff
@@ -68,7 +68,228 @@ namespace :va_facilities do
...
+              facility.tuesday = vaf["Tuesday"]
+              facility.wednesday = vaf["Wednesday"]
+              facility.thursday = vaf["Thursday"]
+              facility.friday = vaf["Friday"]
+              facility.saturday = vaf["Saturday"]
+              facility.sunday = vaf["Sunday"]
+              facility.hours_note = vaf["HoursNote"]
+              facility.save
+              ctr += 1
+              puts 'Updated facility: ' + vaf["Official_Station_Name"] + ': ' + ctr.to_s
```
JoshingYou1, line 225: Can we interpolate here?

lib/tasks/va_facilities.rake
```diff
@@ -68,7 +68,228 @@ namespace :va_facilities do
...
+                  saturday: vaf["Saturday"],
+                  sunday: vaf["Sunday"],
+                  hours_note: vaf["HoursNote"],
+                  hidden: hidden
+              )
+            end
+          else
+            # update record.....
+            facility = VaFacility.where(station_number: vaf["Station_Number"]).first
+            if !facility.blank? && visn.number === vaf["VISN"].to_i
```
JoshingYou1, line 159: Can we do `facility.present?`

JoshingYou1 (changes requested):
Nice work on this Brad! I left a few code comments, but I also wanted to mention that we may want to refactor the `scope` methods in the `va_facility.rb` file to make sure hidden is false there as well.

db/schema.rb
bradjohnson92008: OK... I added it back in.  fyi - I was working on a branch created from master... but since I reviewed the pr that had this change to remove it.. the home page was broke.. since it still had code referencing it.. in the master branch..

lib/tasks/va_facilities.rake
bradjohnson92008: I removed them.

lib/tasks/va_facilities.rake
bradjohnson92008: Ur right - we don't need it.

lib/tasks/va_facilities.rake
bradjohnson92008: Yep  - I'm surprised I didn't have parenthesis... I usually do when I don't need them... and yes I'll just pass true to the create method.

lib/tasks/va_facilities.rake
bradjohnson92008: yep..

lib/tasks/va_facilities.rake
bradjohnson92008: Yep got it.

bradjohnson92008:
I think we only need to add the hidden: false to the these two scope methods:  

scope :get_by_visn, -> (visn) { cached_va_facilities.where(visn: visn, hidden: false) }
scope :get_classification_counts, -> (facility_type) { where(classification: facility_type, hidden: false).size }

lib/tasks/va_facilities.rake
```diff
@@ -68,7 +61,228 @@ namespace :va_facilities do
         end
       end
     end
+    puts "All VA facilities have been created or updated in the DB!"
```
emyl3, line 64: this should just say "All VA facilities have been created in the DB!" since we technically aren't updating anything

lib/tasks/va_facilities.rake
```diff
@@ -68,7 +61,228 @@ namespace :va_facilities do
...
+              puts "Updated facility: #{vaf['Official_Station_Name']}, #{ctr.to_s}"
+            end
+          end
+        end
+      end
+    puts "All VA facilities have been created or updated in the DB!"
+  end
+
+  def create_vet_center_facility(visn, vaf, hidden)
+    VaFacility.create!(
```
emyl3, line 227: for the vet centers, is the only difference between the other facilities (besides the fact that it is a vet center) the `hidden` field being set to `true` or are there other differences? 🤔 Trying to think if we can not repeat some of this. :)

emyl3 (commented):
Left a couple of comments and questions! Thanks, Brad!

lib/tasks/va_facilities.rake
bradjohnson92008: No we are... if the facility exists we update.

lib/tasks/va_facilities.rake
bradjohnson92008: I know... I was trying to do that.. bu there are 7 fields that don't exist in the vet center schema.. and can't add logic within the create statements.  i.e., 
# these 4 not in VetCenter json schema...
                  after_hours_phone_number: vaf["After_Hours_Phone_Number"],
                  pharmacy_phone_number: vaf["Pharmacy_Phone_Number"],
                  enrollment_coordinator_phone_number: vaf["Enrollment_Coordinator_Phone_Number"],
                  patient_advocate_phone_number: vaf["Patient_Advocate_Phone_Number"],

lib/tasks/va_facilities.rake
emyl3: in this particular task we are only creating... I am wondering if we need this task `:create_va_facilities_and_transfer_data` and we can just rename `:update_va_facilities` to be both `:create_or_update_va_facilities` since that's what it is doing

bradjohnson92008: Yeah - the first task 'create_va_facilities_and_transfer_data" is obsolete...  theis is the original task Josh wrote for the first import.  I'll remove that to avoid confusion.  I renamed the task we need to run for this PR to create_or_update_va_facilities.  We should be able to re-use this code for future imports (unless they make changes to the schema of course).

lib/tasks/va_facilities.rake
bradjohnson92008: Actually - you don't even need to do the || nil... :)  if the field doesn't exist the value of vaf["fieldname"] .. is nil... and since we allow nil/null for almost every field in this table... it works.  I just assumed if the field didn't exist... it would throw an error.  So I was able to get rid of that additional method to just create the Vet Centers...

JoshingYou1 (commented):
The only thing left to do is to make sure to remove the old task name and add the new one to the README file. Nice work Brad!

bradjohnson92008:
Good catch Josh... in addition to updating the Readme file I updated the dm.rake task to execute the new task instead of the old one that just created facilities.

lib/tasks/va_facilities.rake
```diff
@@ -1,74 +1,162 @@
...
-  task :create_va_facilities_and_transfer_data => :environment do
-    va_facilities = JSON.parse(File.read("#{Rails.root}/lib/assets/vamc.json"))
+  task :create_or_update_va_facilities => :environment do
+    ctr = 0
+    file = File.read("#{Rails.root}/lib/assets/va_facilities.json")
+    va_facilities = JSON.parse(file)
     visns = Visn.all
-
     visns.each do |visn|
+      puts 'VISN: ' +  visn.number.to_s
```
JoshingYou1, line 10: Do we need this?

JoshingYou1 (commented):
Just left one comment Brad

lib/tasks/va_facilities.rake
bradjohnson92008: I removed it.

emyl3:
on the facilities index page, I am noticing the following:
- "Vet Centers" are appearing in the table
- on first page load the facilities are supposed to be sorted by ABC order by name but they are not
- Facilities with a complexity of 3 have a "3-" 
Screen Shot 2021-09-17 at 10 51 17 AM

on the practice edit page, I am noticing we can select Vet Centers
Screen Shot 2021-09-17 at 10 55 32 AM

not sure if this is an issue but I am able to navigate to individual vet center facility show pages

bradjohnson92008:
Good catch Elisa... those issues are fixed... the va facilities cache was not applying the filter and order.

lib/tasks/va_facilities.rake
```diff
@@ -1,74 +1,160 @@
...
-            wednesday: vaf["Wednesday"],
-            thursday: vaf["Thursday"],
-            friday: vaf["Friday"],
-            saturday: vaf["Saturday"],
-            sunday: vaf["Sunday"],
-            hours_note: vaf["HoursNote"]
-          )
+          if visn.number === vaf["VISN"].to_i && VaFacility.where(station_number: vaf["Station_Number"]).empty?
+            puts 'Creating facility - ' + vaf["Official_Station_Name"]
+              hidden = vaf["CoCClassification"].blank? ? true : false
```
emyl3, line 13: you can do `vaf["CoCClassification"].blank?` and you don't need it to be a ternary

app/views/va_facilities/_index_table_row.html.erb
```diff
@@ -3,7 +3,11 @@
     <th scope="row" role="rowheader" class="grid-col-4"><a class="usa-link" href='<%= va_facility_path(facility) %>'><%= facility["official_station_name"] %> (<%= facility["common_name"] %>)</a></th>
     <td class="grid-col-1"><%= facility.street_address_state %></td>
     <td class="grid-col-1"><a class="usa-link" href='/visns/<%= facility.visn.number %>'><%= facility.visn.number %></a></td>
-    <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level[0..1].upcase %></td>
+    <% if facility.fy17_parent_station_complexity_level.match(/^1/) %>
+      <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level[0..1].upcase %></td>
+    <% else %>
+      <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level[0..0].upcase %></td>
+    <% end %>
```
JoshingYou1, line 10: Can we do a ternary statement here? Something like: 

```
<% facility_complexity = facility.fy17_parent_station_complexity_level %>
<td class="grid-col-2"><%= facility_complexity.match(/^1/) ? facility_complexity[0..1].upcase : facility_complexity[0..0].upcase %></td>
```

JoshingYou1 (commented):
Just left one small nit Brad. Everything else looks great and it all worked as intended for me

app/views/va_facilities/_index_table_row.html.erb
bradjohnson92008: Good call Josh! I fixed it up.

app/models/va_facility.rb
```diff
@@ -8,8 +8,8 @@ class VaFacility < ApplicationRecord
 
   attr_accessor :reset_cached_va_facilities
 
-  scope :get_by_visn, -> (visn) { cached_va_facilities.where(visn: visn) }
-  scope :get_classification_counts, -> (facility_type) { where(classification: facility_type).size }
+  scope :get_by_visn, -> (visn) { cached_va_facilities.where(visn: visn, hidden: false) }
```
emyl3, line 11: we do not need to add the hidden false since it is already being filtered on cached va facilities

emyl3 (changes requested):
left a couple comments! thanks brad!

app/models/va_facility.rb
bradjohnson92008: Removed it...

spec/features/admin/admin_spec.rb
```diff
@@ -574,36 +574,34 @@ def visit_practice_show
...
-    logout(@user2)
-    login_as(@admin, :scope => :user, :run_callbacks => false)
-    visit '/admin/practices/the-best-practice-ever/edit'
-    expect(page).to have_selector('#practice_user_id')
-    fill_in('User email', with: @user.email)
-    click_button('Update Practice')
-
-    expect(Practice.first.commontator_thread.subscribers).to include(@user, @user2)
-  end
+  # it 'if the practice user is changed, it should remove the previous practice user from the comment thread subscribers list for that practice, unless they created at least one comment on the thread' do
```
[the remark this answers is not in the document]
bradjohnson92008: It should still be applicable... I don't think anything changed here.  But I can't get it to work... doesn't see the "User emal" field.   Still trying to fix it.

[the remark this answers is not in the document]
bradjohnson92008: This test is passing now..

spec/features/visn_spec.rb
```diff
@@ -207,7 +207,6 @@ def expect_visn_metadata(element, practices_created_count_text, practices_adopte
     it 'should display a brief breakdown of the visn\'s metadata' do
       visit '/visns/2'
 
-      expect(page).to have_content('This VISN has 3 facilities and serves Veterans in Florida and Georgia.')
```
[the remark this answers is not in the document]
bradjohnson92008: Fixed the sort order - test passes now as is.

app/helpers/visns_helper.rb
```diff
@@ -3,7 +3,7 @@ module VisnsHelper
 
   def get_facility_locations_by_visn(visn)
     sorted_facility_locations = VaFacility.get_by_visn(visn).get_locations
```
emyl3, line 5: nit: you can chain `.sort` onto the end of line 5

emyl3 (approved):
Had one nit but otherwise looks good!

app/helpers/visns_helper.rb
bradjohnson92008: I changed it.

JoshingYou1 (approved):
Awesome job Brad!

PR #484, opened 2022-04-05
Dm 3230

bradjohnson92008:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-3230

Catch all for any  CRH issues that were not implemented in main stories.

1. Navigate to any Visns show page.  i.e., /visns/19
2. Verify that Clinical Resource Hub for that Visn is displayed at the bottom of the 'List of VA facilities in this VISN ' section at the bottom of the page.  i.e.,
image
- Verify that the CRH link works and takes you to the CRH show page for that VISN... i.e., 
image

--Styling Issue...
1.  Test that the styling on the CRH show page is correct and does display the search area if there are 0 created and 0 adopted innovations.  Also verify that the styling is consistent and does not change when..
          a.  There are 0 created and 1 adopted.
          b.  There are 0 adopted and 0 created
          c.  Multiple created and 0 adopted
          d.  Multiple adopted and 0 created.

Finally ensure that the load more button continues to work as intended:  We load 6 Innovation cards at a time.

https://www.figma.com/file/ccwLbt94U9QGfnuFCwOXzY/Final-designs---VA-Diffusion-Marketplace?node-id=103%3A69

## Acceptance criteria
- Make sure we include CRH text/links where necessary:
- VISN show page (list of facilities)
- VISN show page (styling issue when created practices but no adopted practices – see screenshot)
image

## Definition of done


app/controllers/visns_controller.rb
```diff
@@ -73,6 +73,10 @@ def show
 
   def load_facilities_show_rows
     @facilities = VaFacility.get_by_visn(@visn).select(:common_name, :id, :official_station_name, :slug, :station_number, :fy17_parent_station_complexity_level)
+    @clinical_resource_hubs = ClinicalResourceHub.cached_clinical_resource_hubs
+    visn_id = Visn.where(number: params["number"].to_i).first.id
```
JoshingYou1, line 77: I think we can use `.find_by` here. It makes it a little cleaner because you won't have to call `.first` at the end as it only returns one record.

app/controllers/visns_controller.rb
```diff
@@ -73,6 +73,10 @@ def show
...
+    @clinical_resource_hubs = @clinical_resource_hubs.where(visn_id: visn_id)
```
JoshingYou1, line 78: Same here. I think we can use `@visn_clinical_resource_hub` or `@clinical_resource_hub` and `.find_by`, since there's only one `clinical_resource_hub` per visn.

app/views/visns/_show_table_row.html.erb
```diff
@@ -1,10 +1,30 @@
 <% facilities.each do |facility| %>
+  <% path = facility.class.name === "ClinicalResourceHub" ? "/crh/" + facility.visn.number.to_s : va_facility_path(facility)  %>
```
JoshingYou1, line 2: I think you can do `clinical_resource_hub_path(facility)` here

app/views/visns/_show_table_row.html.erb
```diff
@@ -1,10 +1,30 @@
...
-      <%= link_to "#{facility.official_station_name} (#{facility.common_name})", va_facility_path(facility), class: 'usa-link' %>
+      <a class="usa-link" href='<%= path %>'><%= facility["official_station_name"] %>
+        <% if facility.class.name != "ClinicalResourceHub"  %> (<%= facility["common_name"] %>)<% end %></a>
+
+
+      <%# if facility.class.name === "ClinicalResourceHub" %>
+        <%#= link_to "#{facility.official_station_name}", va_facility_path(facility), class: 'usa-link' %>
+      <%# else %>
+        <%#= link_to "#{facility.official_station_name} (#{facility.common_name})", va_facility_path(facility), class: 'usa-link' %>
+      <%# end %>
```
JoshingYou1, line 20: We can remove this if we're not using it

app/views/visns/_show_table_row.html.erb
```diff
@@ -1,10 +1,30 @@
 <% facilities.each do |facility| %>
+  <% path = facility.class.name === "ClinicalResourceHub" ? "/crh/" + facility.visn.number.to_s : va_facility_path(facility)  %>
+  <% if facility.class.name === "ClinicalResourceHub" %>
```
JoshingYou1, line 3: We can create a variable call `is_crh` and set the value to `facility.class.name === "ClinicalResourceHub"`. That way we can just call the variable here and on line `13`

app/views/visns/_show_table_row.html.erb
```diff
@@ -1,10 +1,30 @@
...
+    <% created_count = crh_created_practices_count(facility.visn_id) %>
+    <% adopted_count = crh_adopted_practices_count(facility.visn_id) %>
+  <% else %>
+    <% created_count = facility.practices_created_count %>
+    <% adopted_count = facility.practices_adopted_count %>
```
JoshingYou1, line 8: I think we can use ternary statements here instead of setting the variable twice. We can say something like 
```
created_count = is_crh ? crh_created_practices_count(facility.visn_id) : facility.practices_created_count
```

JoshingYou1 (changes requested):
Nice work on these fixes Brad! I left some comments for ya.

app/controllers/visns_controller.rb
bradjohnson92008: Yep that works.

app/controllers/visns_controller.rb
bradjohnson92008: Problem with find_by is it turns var in to a clinical_resource_hub object versus and active record object... so appending it to the @facilities object doesn't work on the next line.   We can take a further look at this if you want.

app/views/visns/_show_table_row.html.erb
bradjohnson92008: Yep good catch.

app/views/visns/_show_table_row.html.erb
bradjohnson92008: Got it.

app/views/visns/_show_table_row.html.erb
bradjohnson92008: agreed.  :)

app/views/visns/_show_table_row.html.erb
bradjohnson92008: Yes much better...

emyl3 (changes requested):
It looks like we have slightly too much space between the radio buttons and header, as well as the space between the radio button and the search bar when I compare to figma.

In figma, the space between the "Search in this Clinical Resource Hub" and the radio buttons is 16px. In figma, the space between the radio buttons and the search bar looks to be around 40px.
Figma
Screen Shot 2022-04-07 at 6 36 19 PM

this branch
Screen Shot 2022-04-07 at 6 37 58 PM

bradjohnson92008:
Vertical Spacing should be good now between radio buttons and header.

app/controllers/visns_controller.rb
JoshingYou1: So I tried it out, and we could do `@facilities + Array(ClinicalResourceHub.find_by(visn_id: visn_id))`, which may be better performance-wise, since the only reason we are calling a `.where()` here is to turn it into an array.

app/controllers/visns_controller.rb
```diff
@@ -73,6 +73,10 @@ def show
 
   def load_facilities_show_rows
     @facilities = VaFacility.get_by_visn(@visn).select(:common_name, :id, :official_station_name, :slug, :station_number, :fy17_parent_station_complexity_level)
+    @clinical_resource_hubs = ClinicalResourceHub.cached_clinical_resource_hubs
+    visn_id = Visn.find_by(number: params["number"].to_i).id
+    @clinical_resource_hubs = @clinical_resource_hubs.where(visn_id: visn_id)
```
JoshingYou1, line 78: Can we call the second `@clinical_resource_hubs` something else? Maybe like `@visn_clinical_resource_hub`? We're only trying to return one `ClinicalResourceHub`, so I don't think we need it to be plural, plus it's overwriting the instance variable name from line `76`.

JoshingYou1 (commented):
Thank you for making those changes Brad! I responded to your comment about the `find_by` scenario and left one other comment.

app/controllers/visns_controller.rb
bradjohnson92008: Hey Josh - good catch... i'm just using @facilities += Array(ClinicalResourceHub.find_by(visn_id: visn_id)) on line 78 so we don't even need that other variable... and the var on line 76 doesn't get overwritten.

emyl3 (approved):
Looks good to me! 😸 Nice work on this, Brad!

PR #497, opened 2022-05-02
dm-3361

bradjohnson92008:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-3361

Ensures that public users only see created and adopted counts for public innovations on the Facility Index and Show pages.  Internal/authenticated users see counts for both private and public innovations on those pages.

1.  navigate to the Facilities index page as a public user (/facilities).  Note the created and adopted counts for any facility in the table.
image

2. CLick on a facility link to display the facilities show page and ensure that the created and adopted counts are the same on the show page as the facility index page for that facility.
image

4. Sign in and repeat steps 1 and 2.  The adopted and created innovation counts may be greater for the same facility bc now you are seeing both private and public innovations for that facility, but the created, and adopted counts should still be the same on the Facility INdex and show pages.. i.e., 
image
show page:
image

5.  (*bonus) go to the admin panel and mark certain innovations as private or public and verify that the public users do not see counts from private created or adopted practices.

N/A

## Acceptance criteria
Counts for Created and Adopted Innovations may vary between the Facilities Index page and the Facilities show page for “Public Users” … due to innovations being set to “Private” vs “Public”.  Public Users should only see counts based on “Public” innovations.  Internal users should see counts for both public and private ‘innovations.

## Definition of done

### DIFF
--- a/app/controllers/va_facilities_controller.rb
+++ b/app/controllers/va_facilities_controller.rb
@@ -1,5 +1,7 @@
 class VaFacilitiesController < ApplicationController
   include PracticeUtils
+  helper_method :created_practices_count
+  helper_method :adopted_practices_count
   before_action :set_va_facility, only: [:show, :created_practices, :update_practices_adopted_at_facility]
 
   def index
@@ -54,6 +56,24 @@ class VaFacilitiesController < ApplicationController
     created_practices = helpers.is_user_a_guest? ? Practice.get_facility_created_practices(@va_facility.id, nil, 'a_to_z', nil, true) : Practice.get_facility_created_practices(@va_facility.id, nil, 'a_to_z', nil, false)
     @created_pr_count = created_practices.size
     @created_practices_categories = get_categories_by_practices(created_practices, [])
+  end
+
+  def created_practices_count(id)
+    #practices_created_count
+    if helpers.is_user_a_guest?
+      Practice.public_facing.get_by_created_facility(id).size
+    else
+      Practice.published_enabled_approved.get_by_created_facility(id).size
+    end
+  end
+  def adopted_practices_count(id)
+    #practices_adopted_count
+    #Practice.published_enabled_approved.get_by_adopted_facility(id).size
+    if helpers.is_user_a_guest?
+      Practice.public_facing.get_by_adopted_facility(id).size
+    else
+      Practice.published_enabled_approved.get_by_adopted_facility(id).size
+    end
   end
 
   # GET /facilities/:id/created_practices
--- a/app/controllers/va_facilities_controller.rb
+++ b/app/controllers/va_facilities_controller.rb
@@ -1,7 +1,5 @@
 class VaFacilitiesController < ApplicationController
   include PracticeUtils
-  helper_method :created_practices_count
-  helper_method :adopted_practices_count
   before_action :set_va_facility, only: [:show, :created_practices, :update_practices_adopted_at_facility]
 
   def index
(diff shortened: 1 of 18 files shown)


app/controllers/va_facilities_controller.rb
```diff
@@ -1,5 +1,7 @@
 class VaFacilitiesController < ApplicationController
   include PracticeUtils
+  helper_method :created_practices_count
+  helper_method :adopted_practices_count
```
JoshingYou1, line 4: Do we need these? `helper_method` is used for calling a method from a helper module in a controller.

app/controllers/va_facilities_controller.rb
```diff
@@ -56,6 +58,21 @@ def show
...
+    end
+  end
+  def adopted_practices_count(id)
+    if helpers.is_user_a_guest?
+      Practice.public_facing.get_by_adopted_facility(id).size
+    else
+      Practice.published_enabled_approved.get_by_adopted_facility(id).size
+    end
+  end
+
```
JoshingYou1, line 75: I think we can just create an instance variable in whatever method we want to call these in. So something like:

```
@created_practices_count = helpers.is_user_a_guest? ? Practice.public_facing.get_by_created_facility(id).size : Practice.published_enabled_approved.get_by_created_facility(id).size
```

app/views/va_facilities/_index_table_row.html.erb
```diff
@@ -4,7 +4,7 @@
     <td class="grid-col-1"><%= facility.street_address_state %></td>
     <td class="grid-col-1"><a class="usa-link" href='/visns/<%= facility.visn.number %>'><%= facility.visn.number %></a></td>
     <td class="grid-col-2"><%= facility.fy17_parent_station_complexity_level.match(/^1/) ? facility.fy17_parent_station_complexity_level[0..1].upcase : facility.fy17_parent_station_complexity_level[0..0].upcase %></td>
-    <td class="grid-col-2"><%= facility.practices_created_count %></td>
-    <td class="grid-col-2"><%= facility.practices_adopted_count  %></td>
+    <td class="grid-col-2"><%= created_practices_count(facility.id) %></td>
+    <td class="grid-col-2"><%= adopted_practices_count(facility.id)  %></td>
```
JoshingYou1, line 8: We can just call the instance variable here

JoshingYou1 (changes requested):
Nice work Brad! Left a few comments 😃

app/controllers/va_facilities_controller.rb
bradjohnson92008: Yeah - these methods are in the controller and being called from a view (facilities_index table)... bc I can't access the session helper - is_user_a_guest? method in the model - where we were previously getting the counts.

app/controllers/va_facilities_controller.rb
bradjohnson92008: We still have to call the methods for each faciilty.. as the index table is generated..  I can turn the methods in to ternary statements if you prefer.

app/views/va_facilities/_index_table_row.html.erb
bradjohnson92008: The instance var will need to be updated for each row... so we have to call the methods here... we can't just reference the instance var.

app/controllers/va_facilities_controller.rb
JoshingYou1: If that's the case, can we move these methods to a `va_facilities_helper.rb` file? That way we won't need the `helper_method` calls at the top of the class and we can call them in the view whenever we need to.

app/views/va_facilities/_index_table_row.html.erb
JoshingYou1: If that's the case, can we move these methods to a `va_facilities_helper.rb` file? That way we won't need the `helper_method` calls at the top of the class and we can call them in the view whenever we need to.

app/controllers/va_facilities_controller.rb
bradjohnson92008: Yep - that works.... I think I tried the _helper route.. but I must have been calling the "is_user_a_guest?" method wrong... anyway this works and performance is the same.

app/views/va_facilities/_index_table_row.html.erb
bradjohnson92008: Yep that works.  Thx Josh

JoshingYou1 (approved):
Works as intended for me. Nice work Brad!! 😃

PR #506, opened 2022-05-17
Dm 3383

bradjohnson92008:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-3383

Adds a form field to entice bots and other spoof scripts that are maliciously filling out the "Contact us" form on both the "About" and "Nominate" pages.

Makes the about and nominate pages accessible to public users again.

**** run migrations: table: form_spams is created to log suspicious requests.
**** before doing any tests place a debugger in the "Email" method of both the about_controller and the nominate_practices_controller.  In order to inspect the params when the "Contact us" form is submitted.  You may also want to comment out the send_contact_email method just to prevent from accidentally submitting the form and sending emails.

1.  You do not need to be logged in to test this.  *First test is to ensure that the "Contact us" form is accessible to public users from both the "About" link in the top nav bar and the "Nominate an Innovation" from the footer.
2. Click the "About" link from navbar and scroll down to the "Contact us" form.
image
3. Fill in the required fields and click submit. (you should hit the bebugger in the about controller at this point).
4. Inspect the params.  The "phone" param should be blank.  *If the 'phone' param is blank the form will submit and send emails... if it is populated we should log the information from the form to the DB (form_spams table) and redirect the user to the home page.
5.  To test logging of the spam .. populate the 'phone' param with any characters from the debug console i.e., param["phone"] = "*U&Y^T%RR" 
6. Step through the debugger and ensure that the "log_spam_attempt" is called.
7. Ensure that a record is created in the 'form_spams' table for this submit.
image
8.  The "phony" column should contain the chars you populated the "phone" field with.
9. Ensure that the user is redirected to the home page.
10. **Bonus.. ensure that when the "phone" field is blank the send_contact_email method will be called.

*To test the Nominate practices path just navigate to the "Contact us" form by clicking the 'Nominate a practice' link in the footer.  Repeat steps 1-10 above.  Functionality is identical to the About path.

N/A

## Acceptance criteria

- Add hidden fields to nominate and about forms.
- Add logic on submit to verify if hidden field is populated or not.
- If populated: Log info? (ip address(?), form name, url, datetime …etc)
- If not populated submit as normal.

## Definition of done


app/controllers/about_controller.rb
```diff
@@ -3,11 +3,9 @@ def index
   end
 
   def email
-    # TODO remove once we add captcha to form
-    if helpers.is_user_a_guest?
-      respond_to do |format|
-        format.html { redirect_to root_path }
-      end
+    if params["phone"].to_s.length > 0
```
JoshingYou1, line 6: I think we should change the param name to "phony" if that's what we're using for the column in the DB. It may be a little confusing down the line.

app/controllers/application_controller.rb
```diff
@@ -90,4 +90,14 @@ def set_visitor_props
...
     end
   end
+
+  def log_spam_attempt(form)
+    ip_address = request.remote_ip
+    url_orig = request.original_url
+    email = params["email"]
+    subject = params["subject"]
+    message = params["message"]
+    phony = params["phone"]
```
JoshingYou1, line 100: I think we should change the param name to "phony" if that's what we're using for the column in the DB. It may be a little confusing down the line.

app/controllers/nominate_practices_controller.rb
```diff
@@ -1,15 +1,12 @@
...
-    redirect_to root_path if helpers.is_user_a_guest?
   end
 
   def email
-    # TODO remove once we add captcha to form
-    if helpers.is_user_a_guest?
-      respond_to do |format|
-        format.html { redirect_to root_path }
-      end
+    if params["phone"].to_s.length > 0
```
JoshingYou1, line 7: I think we should change the param name to "phony" if that's what we're using for the column in the DB. It may be a little confusing down the line.

app/views/shared/_email_form.html.erb
```diff
@@ -1,9 +1,10 @@
 <%= form_tag "#{path}", method: 'post' do %>
   <label class="usa-label margin-top-4 margin-bottom-2" for="email">Your email address</label>
   <input class="usa-input" id="email" name="email" type="email" required/>
-
   <label class="usa-label margin-top-4 margin-bottom-2" for="subject">Subject line</label>
   <input class="usa-input" id="subject" name="subject" type="text" required/>
+  <label class="usa-label form_h_field margin-top-4 margin-bottom-2" for="phone">Phone</label>
+  <input class="usa-input form_h_field" id="phone" name="phone" type="phone"/>
```
JoshingYou1, line 7: I think we should change the param name to "phony" if that's what we're using for the column in the DB. It may be a little confusing down the line.

spec/features/about_spec.rb
```diff
@@ -57,11 +57,20 @@
...
+      click_button('Send message')
+      message = find('#subject').native.attribute('validationMessage')
+      expect(message).to eq('Please fill out this field.')
+      fill_in('Subject line', with: 'Test subject')
+      fill_in('Your message', with: 'This is a test message')
+      fill_in 'phone', with: 'this is spam'
+      # make sure the mailer count does not increase
+      expect { click_button('Send message') }.to change { ActionMailer::Base.deliveries.count }.by(0)
+      # make sure user is redirected to home page.
+      expect(page).to have_current_path(root_path)
```
JoshingYou1, line 73: Can we add an expectation here to ensure a `FormSpam` record is being created?

spec/features/nominate_practices/nominate_a_practice_page_spec.rb
```diff
@@ -30,16 +30,24 @@
...
+      # all fields should be required
+      click_button('Send message')
+      message = find('#message').native.attribute('validationMessage')
+      expect(message).to eq('Please fill out this field.')
+      fill_in('Your message Please include information about where your innovation is being implemented.', with: 'This is a test message')
+      fill_in('phone', with: 'this is spam')
+      # make sure the mailer count does not increase
+      expect { click_button('Send message') }.to change { ActionMailer::Base.deliveries.count }.by(0)
+      # make sure user is redirected to home page.
+      expect(page).to have_current_path(root_path)
```
JoshingYou1, line 45: Can we add an expectation here to ensure a `FormSpam` record is being created?

JoshingYou1 (changes requested):
Nice work Brad! I left some comments for ya. Just a thought, but shouldn't we put some javascript in place to not submit the form if the "phony" field is filled out? That would help to prevent spam form submissions.

app/views/shared/_email_form.html.erb
bradjohnson92008: I can change the db name to phone...  we want to keep the name 'phone' to look like a legit field.

app/controllers/nominate_practices_controller.rb
bradjohnson92008: I can change the db name to phone... we want to keep the name 'phone' to look like a legit field.

app/controllers/about_controller.rb
bradjohnson92008: I changed the db column to phone to be consistent.  Get latest.. but you probably will need to do a 'rails db:rollback' then rerun the migration.

app/controllers/application_controller.rb
bradjohnson92008: I changed the db column to phone to be consistent. Get latest.. but you probably will need to do a 'rails db:rollback' then rerun the migration.

spec/features/about_spec.rb
bradjohnson92008: added expect expect { click_button('Send message') }.to change(FormSpam, :count).by(1)

spec/features/nominate_practices/nominate_a_practice_page_spec.rb
bradjohnson92008: Added expect: expect { click_button('Send message') }.to change(FormSpam, :count).by(1)

db/schema.rb
```diff
@@ -694,6 +706,7 @@
     t.datetime "session_end_time"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.bigint "process_id"
```
JoshingYou1, line 709: Do we need this?

app/controllers/application_controller.rb
```diff
@@ -90,4 +90,14 @@ def set_visitor_props
...
+
+  def log_spam_attempt(form)
+    ip_address = request.remote_ip
+    url_orig = request.original_url
+    email = params["email"]
+    subject = params["subject"]
+    message = params["message"]
+    phone = params["phone"]
+    FormSpam.create(form: form, original_url: url_orig, ip_address: ip_address, email: email, subject: subject, message: message, phone: phone)
+  end
```
JoshingYou1, line 102: Can we move this to a module? I don't think this method is related to the `ApplicationController`. Also, I'm not sure how this is working because it's a private method.

app/controllers/nominate_practices_controller.rb
JoshingYou1, line 7: I think we can just use `params["phone"].present?` here

JoshingYou1 (changes requested):
Nice work Brad! I just left a few comments. Regarding testing the form submissions, you can just add `visible: false` to the "phone" field and it should work as intended.

db/schema.rb
bradjohnson92008: No I don't think so.. I'll remove that from my local schema file and commit.

app/controllers/application_controller.rb
```diff
@@ -90,4 +90,14 @@ def set_visitor_props
...
   end
+
+  def log_spam_attempt(form)
+    ip_address = request.remote_ip
+    url_orig = request.original_url
+    email = params["email"]
+    subject = params["subject"]
+    message = params["message"]
+    phone = params["phone"]
+    FormSpam.create(form: form, original_url: url_orig, ip_address: ip_address, email: email, subject: subject, message: message, phone: phone)
```
emyl3, line 101: @JoshingYou1 @bradjohnson92008 I'm wondering how useful it would be for us to store potentially malicious user input in our DB? Is it helpful for us to know what the spammer was trying to input or that there was an attempt to input spam? 🤔

If the latter, then I would only save the IP address and request url.

What are your two's thoughts?

app/views/about/index.html.erb
```diff
@@ -116,7 +116,7 @@
     </div>
   </section>
 
-  <% unless is_user_a_guest? %>
+  <%# unless is_user_a_guest? %>
```
emyl3, line 119: nit: can we remove this instead of commenting it out?

emyl3 (changes requested):
Left a question for you all!

app/controllers/application_controller.rb
bradjohnson92008: Thats a good point.. I thought everyone would want more info than less..  but your right... it may just be junk filling up our db...  I could also add a counter column and increment that... if the same IP address hits multiple times.. instead of creating a new record.. just increment the counter.  Whatever you guys think..  I'm just interested to see if it catches anything..  feel like i'm going fishing.. lol.

app/views/about/index.html.erb
bradjohnson92008: Got it.

app/controllers/application_controller.rb
bradjohnson92008: Yeah - I created a form_spams_helper to house the log_spam method and just included in the about and nominate controllers.  I did read a few articles that suggested just putting it in the application controller.. since multiple controllers need to access it.

app/controllers/application_controller.rb
bradjohnson92008: Cool so I removed all the form fields... no reason to store the junk.  most likely if records show up here we'll just need the ip_address and url.

bradjohnson92008:
Added the visible: false to the contact us section for the about and nominate specs.. and tests pass!

JoshingYou1 (approved):
LGTM. Nice work Brad.

emyl3 (approved):
Thanks for your work on this Brad! 😸

PR #551, opened 2022-10-14
DM 3631: Support /communities URL for Community pages

camillevilla:
### JIRA issue link
[DM-3631](https://agile6.atlassian.net/browse/DM-3631?atlOrigin=eyJpIjoiN2ZmMzMxNGViNWZmNGM1MWFlNGY0OGVjNjhiNzY1NzgiLCJwIjoiaiJ9)

- updates the `webdrivers` gem to fix an issue affecting local dev for M1 Macs (https://github.com/titusfortner/webdrivers/issues/237)
- Adds support for `/communities/:page_group/:page_slug/` in the 
- Removes unused subpage breadcrumb code that was creating an extra breadcrumb for this use case
	extra-breadcrumb
- adds a `COMMUNITIES` constant to navigation helper that includes `xr-network`
- Edit breadcrumbs so that a community subpage always routes back to community homepage (e.g. `/communities/xr-network/events` has a breadcrumb back to `/communities/xr-network`)
- Adds a redirect for `/communities`

1. Go to `/admin/page_groups` and create two page groups: `xr-network` and `other`
1. Go to `/admin/pages` and create two subpages: `/xr-network/test` and `/other/test`
1. Go to each page. Neither page should have breadcrumbs.
1. Go to `/communities/xr-network/test`. The content should be the same as `/xr-network/test`.
1. Go to `admin/pages` and create two home pages: `/xr-network/home` and `/other/home`.
1. Go to `/communities/xr-network/test`. The breadcrumb should link to `/communities/xr-network.
1. Go to `/other/test`. The breadcrumb should link to `/other`.
1. Edit `/xr-network/test`. Add an image component. Add a downloadable file component. Add a practice card list component.
1. Go to `/xr-network/test` and click on all page component links to make sure they still work as expected.
1. Go to `/communities`. The page should redirect to `/communities/xr-network`.

## Acceptance criteria

- Determine how URLs are used for other pages
- Determine how many things we can go deep
- Can we do “communities/xr-network/PageName”
- QA URLs
- QA images 
- QA jumplinks

Make sure this doesn't break other pages 
## Definition of done


bradjohnson92008 (approved):
Everything worked for me. Great job on this Camille.

app/helpers/navigation_helper.rb
```diff
@@ -233,16 +234,14 @@ def add_landing_page_breadcrumb(path)
           session[:breadcrumbs] << { 'display': "#{@page.page_group.name}", 'path': path }
         end
 
-        def add_sub_page_breadcrumb
-          session[:breadcrumbs] << { 'display': "#{@page.title}", 'path': @builder_page_path }
-        end
-
```
JoshingYou1, line 239: I think we should keep this code. It was there for instances when the user navigated from a subpage to another page. For example, If the subpage had practices, when they clicked on the card and landed on the practice's show page, they would have a breadcrumb for both the landing page and the subpage

app/helpers/navigation_helper.rb
```diff
@@ -233,16 +234,14 @@ def add_landing_page_breadcrumb(path)
...
         if @page_slug == 'home'
           empty_breadcrumbs
+        elsif COMMUNITIES.include?(@page.page_group.slug) && @builder_landing_page.exists?
+          empty_breadcrumbs
+          add_landing_page_breadcrumb("/communities/#{params[:page_group_friendly_id]}")
         elsif @builder_landing_page.exists?
           empty_breadcrumbs
           add_landing_page_breadcrumb("/#{params[:page_group_friendly_id]}")
-          add_sub_page_breadcrumb
```
JoshingYou1, line 245: Adding to the comment above, we could probably add some code like
```
is_communities_page = COMMUNITIES.include?(@page.page_group.slug)
```
somewhere above the `if/else`, and then do:
```
add_sub_page_breadcrumb unless COMMUNITIES.include?(@page.page_group.slug)
```

config/routes.rb
```diff
@@ -126,4 +126,9 @@
 
   # Custom route for reporting a comment
   # get '/practices/:practice_id/comments/:comment_id/report', action: 'report_comment', controller: 'commontator/comments', as: 'report_comment'
+
+  # Accept pages with /communities prefix, e.g. /communities/xr-network/news
+  get '/communities/:page_group_friendly_id' => "page#show"
+  get '/communities/:page_group_friendly_id/:page_slug' => 'page#show'
+  get '/communities', to: redirect('/communities/xr-network') # temporary redirect until more communities are added
```
JoshingYou1, line 133: I think this is a good work-around! I'm wondering if we should think of a way to make the double-nested routes dynamic for future cases, so we don't have to continually add new routes.

JoshingYou1 (changes requested):
Nice work on this Camille! I left a few comments. I'm also wondering if we should add a redirect in the controller, so if the user tries to visit `xr-network/events`, they get automatically routed to `/communities/xr-network/events`. With this suggestion, we may want to add a constant to the `PageGroup` class, kind of like how we did `PRACTICE_EDITOR_SLUGS`, and just call it `COMMUNITY_SLUGS` or something. That way we could just call it whenever we need to.

config/routes.rb
camillevilla: @JoshingYou1 do you mean Line 133? This route should get removed once we have a second community (and ideally have a Communities home page).

app/helpers/navigation_helper.rb
camillevilla: Sure, I've updated the branch so it retains that method.

camillevilla:
Ok, so far I have
1. removed the commit that touched the subpage breadcrumbs.
2. moved community slugs to a method in the `PageGroup` model.

I'm still researching redirecting users away from `xr-network/events` to `/communities/xr-network/events`.

JoshingYou1 (approved):
Nice refactor Camille! It redirects like a charm now 😄 For V2, we should definitely add a test to cover this scenario.

PR #557, opened 2022-10-26
Add DAP GA analytics snippet to app template

camillevilla:
### JIRA issue link
[DM-3628](https://agile6.atlassian.net/browse/DM-3628)

- Adds an application helper to define which pages should be excluded from Google Analytics
- Adds a snippet to the application view template

TODO:
- [x] test in dev and staging and loop back to DAP program support
- [ ] release to PROD
- [ ] loop back with DAP program for account registration

1. Go to the homepage. Open the inspector and look for the `<!-- We participate in the US government's analytics program. See the data at analytics.usa.gov. -->` comment in `<head>`.

There should be no snippets in the following:
1. a non-public practice
1. practice editor
1. admin pages
1. unpublished PageBuilder pages

## Acceptance criteria
- [ ]

## Definition of done

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -183,4 +183,11 @@ module ApplicationHelper
     end
     location_list
   end
+
+  def ga_private?
+    params["controller"] == "admin" || # all admin pages
+    (params["controller"] == "page" && !@page&.published?) || # unpublished PageBuilder pages
+    (params["controller"] == "practices" && !["show", "index"].include?(params["action"])) || # practice editor pages
+    (params["controller"] == "practices" && params["action"] == "show" && !@practice.is_public) # internal or unpublished practices
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -184,7 +184,7 @@ module ApplicationHelper
     location_list
   end
 
-  def ga_private?
+  def private_path? # https://github.com/digital-analytics-program/gov-wide-code#appropriate-placement
     params["controller"] == "users" || # all user pages
     params["controller"] == "system/status" || # status page
     params["controller"] == "admin" || # all admin pages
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -183,4 +183,11 @@ def get_facility_locations_by_visn(visn)
     end
     location_list
   end
+
+  def ga_private?
```
JoshingYou1, line 187: Nit: I think we can rename this to be a bit more generic, in case we need to use it for something else in the future. Maybe `is_path_private?` or something.

app/helpers/application_helper.rb
```diff
@@ -183,4 +183,11 @@ def get_facility_locations_by_visn(visn)
...
+    params["controller"] == "admin" || # all admin pages
+    (params["controller"] == "page" && !@page&.published?) || # unpublished PageBuilder pages
+    (params["controller"] == "practices" && !["show", "index"].include?(params["action"])) || # practice editor pages
+    (params["controller"] == "practices" && params["action"] == "show" && !@practice.is_public) # internal or unpublished practices
```
JoshingYou1, line 191: I'm wondering if we should add the user pages, like `recommended_for_you`, `users/show`, etc.

JoshingYou1 (commented):
Nice work Camille! Left some comments for ya!

app/helpers/application_helper.rb
JoshingYou1: I would also consider writing a small test for this method 😄

camillevilla: 👍 I ended up writing a feature test for this instead of unit tests. 

🤔 I hear what you're saying about potentially reusing this method down the line. I'd be nervous about someone accidentally reusing this in a view where authentication should be done in the controller instead. I added a little comment to the method that links to the DAP implementation page.

bradjohnson92008 (approved):
Great job Camille - LGTM as well.

PR #602, opened 2023-05-23
Move rails-assets gems to a source block

camillevilla:
## Description - what does this code do?
Address deprecation warning: 
```
[DEPRECATED] Your Gemfile contains multiple primary sources. Using `source` more than once without a block is a security risk, and may result in installing unexpected gems. To resolve this warning, use a block to indicate which gems should come from the secondary source.
```

## Acceptance criteria
- [ ]

## Definition of done
- [ ] Unit tests written (if applicable)
- [ ] e2e/accessibility tests written (if applicable)
- [ ] Events are logged appropriately
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating JIRA issue
- [x] No sensitive information (i.e. PII/credentials/internal URLs/etc.) is captured in logging, hardcoded, or specs

PR #613, opened 2023-06-21
Force rails-assets-normalize-css to load from railsassets.org

camillevilla:
- Continues work from #602, where we addressed a deprecation warning in the Gemfile. We are now explicitly declaring any dependencies that are not loaded from rubygems.org. Previously, the application would look for dependencies in rails-assets.org if they were not found on rubygems. 
- addresses error from deployment:
```
Your bundle is locked to rails-assets-normalize-css (3.0.3) from rubygems
repository https://rubygems.org/ or installed locally, but that version can no
longer be found in that source. That means the author of
rails-assets-normalize-css (3.0.3) has removed it. You'll need to update your
bundle to a version other than rails-assets-normalize-css (3.0.3) that hasn't
been removed in order to install.
```
- `rails-assets-normalize-css` is an dependency of a dependency (`uswds-rails`) that was implicitly being loaded from rails-assets.org. In the Gemfile we are explicitly telling bundler to rails-assets.org. We could also fix this in the upstream gem fork [agilesix/uswds-rails](https://github.com/agilesix/uswds-rails), but we'll be moving away from this fork soon anyway.

- deployed this branch to STG and the app built successfully. Styling and JS interactions are as expected.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -126,6 +126,7 @@ source 'https://rails-assets.org' do
   gem 'rails-assets-jquery'
   gem 'rails-assets-sticky'
   gem 'rails-assets-jquery.scrollTo'
+  gem 'rails-assets-normalize-css' # force dependency for uswds-rails to load from rails-assets
 end
 gem "nested_form"
 gem 'colorize'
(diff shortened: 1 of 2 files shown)


bradjohnson92008 (approved):
Thanks for resolving this... great work Camille!

PR #661, opened 2023-08-04
Update Event validation method and unit tests

camillevilla:
### JIRA issue link
Part of DM-3780 efforts

- Updates a validation on PageEventComponent that was preventing users from creating new events and causing `paginated_components_spec.rb` to fail locally and in CI. 

1. Log in as an admin and access a PageBuilder page.
2. Add a new event component with no date information
3. Save

Your new component should be saved.


PhilipDeFraties (approved):
looks good

PR #712, opened 2023-09-28
Remove JS for "Return to top" links and replace with HTML-based implementation

camillevilla:
### JIRA issue link
N/A

- remove JS used in "Return to top" links
- implement "Return to top" as anchor links 

1. Open the homepage at a mobile width. Click on "Return to top" link. The browser should scroll to the top of the page
2. Go to any innovation and click on the "Edit innovation" button. 
3. Click on the "Return to top" link. The browser should scroll to the top of the page.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -36,7 +36,6 @@
 //= require timeago
 //= require replies
 //= require gmaps/google
-//= require return-to-top
 //= require comment
 //= require richmarker
 //= require infowindow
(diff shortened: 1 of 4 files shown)


PhilipDeFraties (commented):
Code looks good and the return to top buttons work, one thing I am curious about though.. 

On the innovation editor pages if there is any text or selections made in the forms when the button is clicked then that information is lost when the page reloads. I checked against the current logic in the master branch and the button does scroll to the top of the page but it doesn't lose any text or selections that have been entered because the page does not reload. 

It seems that this actually reduces functionality a small bit, are we sure this would be a wanted change?

camillevilla:
Alack! I'll scrap this one for now.

PR #725, opened 2023-10-19
DM-4258: Improve fallback behavior for innovation show page images

camillevilla:
### JIRA issue link
[DM-4258](https://agile6.atlassian.net/browse/DM-4258)

- This PR improves the fallback behavior for innovation show pages where image requests fail. Previously, captions would be set to 50px wide columns when S3 image requests failed. This PR adds a conditional so that broken image captions appear at full column width instead. 

As an innovation editor:
1. upload an image that is 50px wide. Give it a caption that is about 190 characters long. 
2. upload an image that is 500px wide. Give it a caption that is about 190 characters long. 
3. Save changes
4. Go to the innovation page
5. Confirm that the 500px wide image caption is as wide as the image
6. Confirm that the narrow image's caption is 450px wide.

### Before
image003

### After

Screenshot 2023-10-19 at 4 17 46 PM

### DIFF
--- a/app/assets/javascripts/practice_page.es6
+++ b/app/assets/javascripts/practice_page.es6
@@ -109,7 +109,11 @@
             $('.practice-editor-impact-photo').each(function () {
                 $(this).parent().show();
                 if ($(window).width() > 1023) {
-                    $(this).next().width(`${$(this).width()}`);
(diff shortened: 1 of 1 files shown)


PhilipDeFraties (approved):
lgtm!

PR #731, opened 2023-11-07
dm-4093 add secondary practice emails to comment notifications

PhilipDeFraties:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-4093

* Adds `practice.practice_emails` to innovation comment email notifications
* refactors `Commontator::SubscriptionsMailer` and `Commontator::Subscription`
* updates `spec/features/practice_viewer/contact_spec.rb`, adds tests, refactors file to use factory bot
* adds factory for `practice_emails`
* updates comment email notification template for when the comment is a reply to another comment (2nd image in screenshots)

* Note: I noticed that if a user comments on the innovation page, that user is from then on considered a "subscriber" to the entire thread and will be cc'd in all future comment notification emails. This seems like it maybe is not intentional, if I were a user that left a singular comment on an innovation's page I would not appreciate being emailed every time someone else comments on the page especially since there is no "unsubscribe" option in the email from what I can tell 🤷🏼

* Note follow-up: The issue of all thread "subscribers" being emailed for every comment has been fixed, now the only additional user that will be included in a new comment email notification will be the creator of a parent comment when the new comment is a reply. 

1. Edit a practice and add a few "Copied" email addresses at the bottom of the `about` page:

Screenshot 2023-11-07 at 4 37 32 PM

2. Verify that the practice's `support_network_email` is different than the practice's `user.email`
3. Log in as a user that is unaffiliated with the practice
4. Go to the innovation show page and create a comment
5. Verify that an email renders (opens in a new window) and that it is addressed to the practice's `user.email`, `support_network_email`, and the email addresses you added as the "Copied" on the about page.
6. Verify that the body of the email is in the expected format with a link to the innovation show page (1st screenshot below)
7. Log out and log in as a different user, comment on the previous comment
8. Verify another notification email is generated and includes the email of the creator of the parent comment and that the body indicates the new comment is a reply to an existing comment (see 2nd screenshot below)

Screenshot 2023-11-08 at 3 53 56 PM

Screenshot 2023-11-08 at 3 48 51 PM

## Acceptance criteria
- [ ]

## Definition of done


app/mailers/commontator/subscriptions_mailer.rb
```diff
@@ -12,39 +12,47 @@ def comment_created(comment, recipients)
...
-    comment_user = User.find(comment.creator_id).full_name
-    @comment_header = "#{comment_user} has commented on #{@comment.thread.commontable.name}"
-    @comment_header = "#{comment_user} has replied to a comment on #{@comment.thread.commontable.name}" unless @comment.parent_id.nil?
-    @thread = @comment.thread
-    @creator = @comment.creator
-    @mail_params = { from: @thread.config.email_from_proc.call(@thread) }
-    practice = Practice.find(@comment.thread.commontable_id)
-    support_network_email = practice.support_network_email || nil
-    @recipient_emails = recipients.map do |recipient|
-      support_network_email.present? && recipient === support_network_email.downcase ? recipient : Commontator.commontator_email(recipient, self)
```
PhilipDeFraties, line 24: I opted to remove `Commontator.commontator_email` because all it actually does it retrieve the user's email address: https://github.com/department-of-veterans-affairs/diffusion-marketplace/blob/4ddbe04e5ad7fd97bb44a5a11e252bb36adf2538/config/initializers/commontator.rb#L78

spec/features/practice_viewer/contact_spec.rb
camillevilla, line 1: 👏 Good job DRYing out this spec file a bit

app/mailers/commontator/subscriptions_mailer.rb
```diff
@@ -12,39 +12,48 @@ def comment_created(comment, recipients)
...
-    end
-    @using_mailgun = Rails.application.config.action_mailer.delivery_method == :mailgun
```
camillevilla, line 26: This param seems like it's still being used in `#setup_mail_params`?

camillevilla (changes requested):
@PhilipDeFraties good catch re: comments going out to all former commenters. Narrowing the scope of these emails is definitely an improvement to user experience. 

RE: removing `@using_mailgun`. I think we should test this PR out in DEV to confirm whether mailgun is actually used in our prod setup. It's not explicitly set in our `production.rb`. We wouldn't be able to catch this locally because local dev uses `letter_opener`. If we're not using mailgun, we can remove the ternary operators from `@mail_params`.
https://github.com/department-of-veterans-affairs/diffusion-marketplace/blob/9de063aa732aa28a9f2e9cc0484f3772a60eab27/config/environments/development.rb#L34

app/mailers/commontator/subscriptions_mailer.rb
PhilipDeFraties: good catch! I forgot to follow up on the removal of this

nuel247 (approved):
We can merge, reviewed and approved

PR #786, opened 2024-02-13
DM-4435: Handle categories with ampersands in homepage search dropdown

camillevilla:
### JIRA issue link
[DM-4435](https://agile6.atlassian.net/browse/DM-4435)

## Description - what does this code do?
- Updates encoding to ensure categories with ampersands (e.g. `Obstetrics & Gynecology`, `Nutrition & Food`) both 
  - render correctly in the dropdown
  - are handed off to search

### Before
1. If you're testing in local dev, log in as an admin. 
1. In `feature/homepage-refresh`, search for `Obstetrics` or `Nutrition`.
1. Categories with ampersands should render with `u\2006`
1. Select `Nutrition \u2006 Food` and submit search
1. No search results should be returned
1. Reset the filters on /search 
1. Select Nutrition & Food from the filters and Update results
1. You should get 8 results back

### After
1. In `feature/homepage-refresh`, search for `Obstetrics` or `Nutrition`.
1. Categories with ampersands should render the character correctly
1. Select `Nutrition & Food` and submit search
1. You should get 8 results back

### Before
Screenshot 2024-02-08 at 11 10 59 AM

### After
Screenshot 2024-02-13 at 10 13 28 AM

PR #787, opened 2024-02-14
Add innovations to homepage search dropdown

camillevilla:
### JIRA issue link
[DM-4431](https://agile6.atlassian.net/browse/DM-4431)

- Adds Innovations to search dropdown
- Only shows public innovation to public users
- Cleans up markup and CSS to make it reusable for additional dropdown sections
- Adds responsive truncation to long result names
- Updates placeholder text [per team feedback](https://agile6.atlassian.net/browse/DM-4431?focusedCommentId=11915)
- Backfills some tests for #786 #779

In local dev:
1. Go to homepage and confirm you only see public innovations
2. Sign in and then return to the homepage
3. You should see 3 innovations, ordered by their `created_at` date
4. Start typing a string like `service`. As you type you should see both Innovations and Categories updating
5. Click on an innovation and confirm that it takes you to the innovation show page

### Public user (local dev)
Screenshot 2024-02-20 at 4 02 33 PM

### Private user (local dev)
Screenshot 2024-02-20 at 4 02 41 PM

https://www.figma.com/file/D16fyBygVx88HpyczTuqXs/Diffusion-Marketplace-(2024)%2B?type=design&node-id=3%3A5426&mode=design&t=P8YnvwIpLSCrNRXh-1

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -7,6 +7,7 @@ class HomeController < ApplicationController
     @highlighted_pr = Practice.where(highlight: true, published: true, enabled: true, approved: true).first
     @category_names_by_popularity = get_category_names_by_popularity
     @featured_topic = Topic.find_by(featured: true)
+    @practice_names = get_practice_names
   end
 
   def diffusion_map
@@ -74,6 +75,22 @@ class HomeController < ApplicationController
   end
 
   private
+
+  def get_practice_names
+    if current_user
+      all_current_practices
+    else
+      all_public_practices
+    end
+  end
+
+  def all_current_practices
+    Practice.where(published: true, retired: false).pluck("name")
+  end
+
+  def all_public_practices
+    Practice.where(published: true, is_public: true, retired: false).pluck("name")
+  end
 
   def get_diffusion_histories(is_public_practice)
     DiffusionHistory.get_with_practices(is_public_practice).order(Arel.sql("lower(practices.name)"))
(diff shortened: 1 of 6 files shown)


app/controllers/home_controller.rb
```diff
@@ -75,6 +77,24 @@ def pii_phi_information
...
+    return practices_and_slugs
+  end
+
+  def all_current_practices
+    Practice.where(published: true, retired: false).order("created_at DESC")
+  end
+
+  def all_public_practices
+    Practice.where(published: true, is_public: true, retired: false).order("created_at DESC")
+  end
```
PhilipDeFraties, line 96: this works well, but at the risk of getting a bit code-golf-y we could potentially DRY things up just a bit:

```suggestion
 def practice_names_and_slugs_hash
  practices = current_user ? all_practices(current: true) : all_practices(public: true)
  practices.pluck("name", "slug").each_with_object({}) do |(name, slug), hash|
    hash[name] = slug
  end
end

def all_practices(current: false, public: false)
  scope = Practice.where(published: true, retired: false)
  scope = scope.where(is_public: true) if public
  scope = scope.order("created_at DESC")
  scope
end
```

since we are not utilizing the return value of the `map`s, I think it's best practice to use `each`, but then we can take it just a step further and use `each_with_object` with a hash to reduce repeated code and save a few lines. 

Not super necessary if you feel strongly that yours reads better 👍🏼

PhilipDeFraties (approved):
Other than my one optional suggestion this works well and lgtm!

PR #793, opened 2024-02-27
DM-4438: Fix dropdown backspacing bug

camillevilla:
### JIRA issue link
[DM-4438](https://agile6.atlassian.net/browse/DM-4438)

- Fixes a bug where entering text into the dropdown and deleting it returned only 2 results and a blank entry in between. 

To reproduce error (in STG or a local dev prior commit): 
1. Click on the search bar. You should see 3 most popular category results and 3 most recent innovation results.
2. Start typing in the search bar, e.g. cardiology.
3. You should see the results eventually narrow down to 1 or fewer results for each section
4. Backspace until there are no characters in the search bar
5. You will only see 2 of the original popular categories and recent innovations. 

To confirm fix:
1. Click or tab into the dropdown. You should see 3 results each for categories and innovations
2. Start typing and notice results change
3. Backspace until there are no characters in the search bar
4. Confirm the 3 results for categories and innovations match the list on the first dropdown init

### Before
backspace-search (1)

### After
after-dropdown-bug


app/assets/javascripts/search.es6
```diff
@@ -17,11 +17,12 @@ function setupSearchDropdown(formId) {
 
     const allCategoriesString = $('.homepage-search').attr('data-categories');
     const allCategories = allCategoriesString ? allCategoriesString.match(/[^",\[\]]+/g) : [];
-    const mostPopularCategories = allCategories ? allCategories.slice(0, 3) : [];
+    const mostPopularCategories = allCategories ? allCategories.slice(0, 5).filter( str => str.startsWith(" ") == false ) : [];
 
     const allInnovationsString = $('.homepage-search').attr('data-innovations');
     const allInnovations = allInnovationsString ? allInnovationsString.match(/[^",\[\]]+/g) : [];
```
PhilipDeFraties, line 23:
```suggestion
    const allInnovations = (allInnovationsString !== "[]") ? allInnovationsString.match(/[^",\[\]]+/g) : [];
```

app/assets/javascripts/search.es6
```diff
@@ -17,11 +17,12 @@ function setupSearchDropdown(formId) {
 
     const allCategoriesString = $('.homepage-search').attr('data-categories');
     const allCategories = allCategoriesString ? allCategoriesString.match(/[^",\[\]]+/g) : [];
```
PhilipDeFraties, line 19:
```suggestion
    const allCategories = (allCategoriesString !== "[]") ? allCategoriesString.match(/[^",\[\]]+/g) : [];
```

PhilipDeFraties: looks like I introduced a bug with my initial code here, when `allCategoriesString` returns `[]` it causes an issue when trying to run the regex on the next line. I didn't realize because Categories was never empty, but now with the addition of Innovations that can be empty the issue has revealed itself.. 

Upon initial testing it looks like this small change will act as a guard and prevents errors.

PhilipDeFraties (approved):
Works great, thanks for fixing!

PR #850, opened 2024-04-16
Remove feature branch from codeQL workflow

camillevilla:
`feature/homepage-refresh` was merged on 3/20 #771


PhilipDeFraties (approved):
I completely forgot I added this, which turned out wasn't even necessary to begin with, thanks for catching!

PR #923, opened 2024-07-01
dm-4933 load gmaps javascript libraries async

PhilipDeFraties:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-4933

Adjusts loading script for gmaps javascript libraries to be async. 

This is something that I started with when working on improving page load times because it was indicated as being necessary as per warning in the devtools console upon page load. In the warning (see screenshot 1) it indicates this might lead to performance gains, however I don't think this is the case judging by what I saw on DEV with this branch deployed. I do still think it is a worthwhile change as I went on to implement changes to the backend and found that the feature tests involved required this change in order to pass, so we will need to merge this branch and then rebase the other branches involved in the story.

I ran into issues with the loading for the libraries happening out of order and redundantly after already being loaded, so I opted to a script file that handles the loading both async and sequentially by loading the main library first then each subsequent library within the callback functions. I would have liked (and tried to) abstract this script to a js file but it needs to leverage the gmaps API key which we cannot expose to clients so the script exists within the `<body>` tag of the `application.html.erb` file.

Due to the nature of the new library loading strategy for gmaps it also took care of the other deprecation warning appearing in the console relating to the use of `google.maps.event.addDomListener()`.

Note: My initial tests don't show any discernible gains in the page load times so I don't think it is beneficial to benchmark.

1. Both locally and with branch deployed to DEV verify the map loads for all views that utilize gmaps, so `/diffusion-map`, page-builder, and innovation pages.
2. Verify the error from the first screenshot below doesn't appear in the devtools console.

before:
Screenshot 2024-07-08 at 10 41 48 AM

after:
Screenshot 2024-07-08 at 10 41 16 AM

## Acceptance criteria
- [ ]

## Definition of done


camillevilla (approved):
Great work! Definitely like the consolidated script loading.

PR #926, opened 2024-07-02
dm-4931 cache practices json data

PhilipDeFraties:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-4931

Adds caching for individual practice's return value for `as_json` which is run in the request response for `/diffusion-map` page as well as the `/search` page.

Invalidation of the cached values is handled via `after_commit` callback `check_update_to_clear_cache`, which replaces `clear_searchable_cache_on_save` which was utilized by a combination of `before_save` and `after_save` callbacks in order to clear the given record's cached values when specific attributes were updated. The `clear_searchable_cache` method was also added as an `after_commit` callback on the related associations' models which similarly replaced existing callbacks.

Note: I'll do the benchmarking, for review purposes just # 1 is fine:
1. Locally with redis running, verify the /diffusion_map page loads, reload a few times.
2. On DEV, before deploying, record the rough average of FCP, LCP times.
3. With the branch deployed to DEV, record load times again and compare.

## Acceptance criteria
- [ ]

## Definition of done
- [x] Events are logged appropriately


app/models/practice.rb
```diff
@@ -52,39 +50,19 @@ def self.cached_published_enabled_approved_practices
...
-        self.overview_problem_changed? ||
-        self.overview_solution_changed? ||
-        self.overview_results_changed?  ||
-        self.retired_changed? ||
-        self.retired_reason_changed? ||
-        self.hidden_changed? ||
-        self.is_public_changed?
-      self.reset_searchable_cache = true
-    end
-  end
```
camillevilla, line 82: Let's dig in to the git blame and history for this a bit before removing:
https://github.com/department-of-veterans-affairs/diffusion-marketplace/blame/master/app/models/practice.rb

PhilipDeFraties: @camillevilla I'm not really seeing any indication as to why Aurora went with this method, but I think that since all the values checked represent attributes on the the table it's a safe bet that utilizing `after_commit` will achieve the same end result. 

The only difference will be that it will now clear the cache if any attribute is updated, rather than just this selection (`practices` has a lot of columns). But I think that the slight performance lost from having an Innovation's cache cleared a bit more often will be negligible and that the reduction in complexity is worth it, but that is up for debate of course.

spec/models/practice_spec.rb
```diff
@@ -165,45 +165,23 @@
...
+  describe '#check_update_to_clear_cache' do
+    let(:practice) { create(:practice) }
 
-        expect(result).to include(practice1)
-        expect(result).not_to include(practice2, practice3)
+    context 'when important attributes change' do
+      it 'clears the cache' do
+        allow(practice).to receive(:clear_searchable_cache)
+        practice.update(name: 'New Name')
+        expect(practice).to have_received(:clear_searchable_cache)
```
camillevilla, line 175: Nice. I like having the intention explicitly spelled out in the spec here

app/models/practice.rb
```diff
@@ -51,38 +49,31 @@ def self.cached_published_enabled_approved_practices
...
-        self.overview_problem_changed? ||
-        self.overview_solution_changed? ||
-        self.overview_results_changed?  ||
-        self.retired_changed? ||
-        self.retired_reason_changed? ||
-        self.hidden_changed? ||
-        self.is_public_changed?
-      self.reset_searchable_cache = true
+    if saved_changes.keys.any? { |key| important_attrs.include?(key) }
+      self.clear_searchable_cache
```
camillevilla, line 61: 🧹 Nice. Good DRY-up here!

camillevilla (approved):

PR #948, opened 2024-07-17
Retire badges 🧹

camillevilla:
### JIRA issue link
N/A

- Badges have not been used in the application since 2020. Let's remove the associated tasks from the local dev importer + the associated DB columns and models.

-Drop the DB and run the `rake dm:full_import` task to confirm the importer task is still working as expected.


db/migrate/20200402174615_add_shark_tank_badge_to_badges.rb
camillevilla, line 1: This and the GoFish migration had to be deleted because they use ActiveRecord and reference models that no longer exist.

nuel247 (approved):
Thanks Camille!!! This is good to go.

PhilipDeFraties (approved):

PR #972, opened 2024-08-08
Dm 4913 new search results

PhilipDeFraties:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-4913

Adds new filter and search results structure to search page

1. Locally, visit the search page and verify the filters section appears as per screenshot-1
2. Verify that each filter works as expected, keep in mind that the filtering logic includes innovations that are true for any filter param selected (does not have to match all)
3. Verify that when resizing the screen to less than mobile width the filters appear under an accordion, with all the selections previously applied in place, verify the reverse when enlarging to desktop view.

Screenshot 2024-08-12 at 3 37 01 PM

Note: the mockups are a bit dated as there have been discussed and approved changes, such as the sort section being changed back to radio buttons and moved to the top of the the "Refine results" section. 

Also, in the mocks there is a header above the results list, I asked for clarification on whether this should be dynamic since the mock has the text as "All Innovations" which seems inaccurate if they are being filtered, Nick was too busy at the time to confirm what is expected there so I've left it out for now until we get closer to roll-out when the team has a chance to review the page on STG.

https://www.figma.com/design/D16fyBygVx88HpyczTuqXs/Diffusion-Marketplace-(2024)%2B?node-id=2109-8042&t=r0WgKSLPhVPW47hS-1

## Acceptance criteria
- [ ]

## Definition of done


app/views/practices/_search.js.erb
```diff
@@ -257,10 +242,10 @@ function searchPracticesPage() {
             // Print results to the page
             document.querySelector('#search-results').innerHTML = finalResults;
             // After the html for the card has been built and applied to the DOM, replace the image 'src' based on the signed url
-            replaceImagePlaceholders();
+            replaceSearchPageImagePlaceholders(practiceElements);
             $(SEARCH_RESULTS).showMoreItems({
-                startNum: 12,
-                afterNum: 12,
+                startNum: 24,
+                afterNum: 24,
```
camillevilla, line 248: Interesting - what's the threshold change doing here?

app/views/practices/_search.js.erb
```diff
@@ -1055,9 +998,6 @@ function searchPracticesPage() {
 
         // Move the focus area back to the top of the filters window
         $(".usa-combo-box__clear-input__wrapper").attr("tabindex",-1).focus();
```
camillevilla, line 1000: This is a leftover from when the clear filters button was below the accordion. I think it makes sense to either remove this or rejigger where the focus is sent. Also remove the reference to `#resetSearchFiltersButtonMobile` up above now that related ID is gone.

app/views/practices/search_partials/_search_filters.html.erb
```diff
@@ -1,85 +1,93 @@
...
+      ['Sort by most recently added', 'added']
+    ]
+  %>
+  <h4 class="text-bold">Sort By</h3>
+  <% sort_options.each do |option| %>
+    <div class="usa-radio">
+      <input id="search_sort_option_<%= option[1] %>" class="usa-radio__input" type="radio" name="search_sort_option" value="<%= option[1] %>" <%= 'checked' if option[1] == 'adoptions' %> aria-label="<%= option[0] %>">
+      <label class="usa-radio__label" for="search_sort_option_<%= option[1] %>" aria-label="<%= option[0] %>">
+        <%= option[0] %>
+      </label>
```
camillevilla, line 18:
```suggestion
      <input id="search_sort_option_<%= option[1] %>" class="usa-radio__input" type="radio" name="search_sort_option" value="<%= option[1] %>" <%= 'checked' if option[1] == 'adoptions' %>>
      <label class="usa-radio__label" for="search_sort_option_<%= option[1] %>">
        <%= option[0] %>
      </label>
```
The aria-labels here can be removed since they're duplicating content. Also consider using `fieldset` and `legend` here to get the radio buttons semantically grouped nicely: https://designsystem.digital.gov/components/radio-buttons/

app/views/practices/_search.js.erb
PhilipDeFraties: I had to add divider elements and those elements are added dynamically depending on the index of the search result above it, if the search result is the last one it doesn't add a divider

app/assets/javascripts/_practice_card_utilities.es6
```diff
@@ -74,6 +72,47 @@ function replacePlaceholderWithImage(imageUrl, practiceId, practiceName) {
...
+    });
+}
+
+function replaceSearchResultPlaceholderImage(imageUrl, practiceId, practiceName) {
+    return new Promise((resolve, reject) => {
+        loadImage(imageUrl, function(loadedImageSrc) {
+            const imgElement = $('<img>')
+                .attr('data-resource-id', practiceId)
+                .attr('src', loadedImageSrc)
+                .attr('alt', practiceName + ' Search Result Image')
```
camillevilla, line 105: I'd suggest we set these as `alt=""` if there's no alt text available in `practice.main_display_image_alt_text`. I know this is pre-existing behavior, but I don't think it's worth perpetuating.

camillevilla, line 105:
```suggestion
```
```suggestion
```
I'd suggest we set these as `alt=""` if there's no alt text available in `practice.main_display_image_alt_text`. This creates extra noise for screenreader users without conveying useful info about the image contents. I know this is pre-existing behavior, but I don't think it's worth perpetuating.

app/views/practices/_search.js.erb
PhilipDeFraties: @camillevilla I just realized that this wasn't exactly correct, what I needed to do was add the lines above the search result elements (when not the first) and set the start and after to 23, which is more in line with the design

app/views/practices/search_partials/_search_filters.html.erb
PhilipDeFraties: @camillevilla I actually consciously chose not to make it a fieldset because from what I read a fieldset is intended to be part of a form, which the sort options are not, is it ok to make them a fieldset you think?

app/views/practices/_search.js.erb
camillevilla: 🤔 Hmm if we're needing to recalculate pagination due to adding `<hr>` elements, I'd suggest we implement those as a `border-bottom` via CSS instead. Added bonus of avoiding `<hr>` elements - we don't have to remember to add an `aria-hidden="true"` to them.

app/views/practices/search_partials/_search_filters.html.erb
camillevilla: @PhilipDeFraties ahh ok I see what you're saying re: not-form-ness here. The `Sort by` options immediately update the results. In that case I'm thinking we might need to add some `aria` to indicate that results immediately changed...and... yeah let's do that in another PR discussion.

camillevilla (approved):

camillevilla (approved):

camillevilla:
This looks good to me. Let's merge this into `feature`.

PR #1010, opened 2024-09-04
dm-5067 product active admin

PhilipDeFraties:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-5067

Adds `active-admin` interface for creating, updating, and viewing `products` from the admin panel

Largely sticks with the patterns established in the `admin/practices` tab but with only bare-necessary attrs for now (`name` and `user`).

1. Locally go to the `admin/products` tab and click the "New Product" button to go to the product creation form
2. Create a product with a name but no "USER EMAIL", check the "Create another Product" box before submitting
3. Verify the new product form renders and the successful creation flash message appears
4. Try to create another product with the same name as the last, verify a failure flash message appears indicating the validation failure, submit with a different name and verify the product is created.
5. Verify a product can be created with an existing user's email for the "USER EMAIL" field, verify that a non-VA email prevents creation, repeat for updating a product.
6. Verify that when creating and updating with a valid VA email that doesn't belong to a current user the creation is successful and a new user has been created and persisted to the db with the submitted email (this is an established pattern in `admin/practices`)

Screenshot 2024-09-04 at 12 34 47 PM
Screenshot 2024-09-04 at 12 35 01 PM
Screenshot 2024-09-04 at 12 35 13 PM
Screenshot 2024-09-04 at 12 35 27 PM

## Acceptance criteria
- [ ]

## Definition of done


app/admin/products.rb
```diff
@@ -0,0 +1,111 @@
...
+
+ActiveAdmin.register Product do
+  config.create_another = true
+  actions :all, except: [:destroy]
+  permit_params :id, :user_email, :create_another
+
+  # To do: set up index file export
+  index download_links: false do
+    id_column
+    column 'Product Name', :name
```
camillevilla, line 11:
```suggestion
    column 'Owner email', :user_email
    column 'Created at', :created_at
    column 'Updated at', :updated_at
    actions
```
This is a little too minimal right now. Let's add more information to the index and the basic set of actions.

app/admin/products.rb
```diff
@@ -5,7 +5,8 @@
   actions :all, except: [:destroy]
   permit_params :id, :user_email, :create_another
 
-  index do
+  # To do: set up index file export
+  index download_links: false do
```
camillevilla, line 9:
```suggestion
  index do
```

camillevilla: Let's keep the CSV export in place and disable fields later. We can use this as a way to quickly QA the bulk import workflow.

app/admin/products.rb
```diff
@@ -0,0 +1,111 @@
+include UserUtils
+
+ActiveAdmin.register Product do
+  config.create_another = true
+  actions :all, except: [:destroy]
```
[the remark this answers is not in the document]
camillevilla: I would suggest keeping `:destroy` in place until we can iron out the intended Publish / Retire / Disable workflow with the team.

camillevilla (changes requested):

camillevilla (changes requested):

app/admin/products.rb
```diff
@@ -0,0 +1,119 @@
+include UserUtils
+
+ActiveAdmin.register Product do
+  config.create_another = true
+  actions :all
+  permit_params :id, :user_email, :create_another
```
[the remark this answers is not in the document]
PhilipDeFraties: so after doing some digging I discovered that using `permit_params` is problematic because it is intended to only be used with params that pertain directly to the model's attributes. `user_email` was throwing a warning in the console as being disallowed so I opted for 2 traditional strong params methods. It may seem a bit verbose but `active_admin` was meant to be more streamlined and less customized than we are going for, so in order to keep things safe and conventional in regards to params I think this is our best bet.

camillevilla (approved):

PR #1042, opened 2024-09-23
dm-5079 product editor multimedia page

PhilipDeFraties:
### JIRA issue link
https://agile6.atlassian.net/browse/DM-5079

adds `product/edit/multimedia` page for adding and managing of `product` images, videos, links, and video links

note: the page layout is a bit different than the design, this is because I reused much of the hmtl and js that provides the resource management ui for the "practice" editor. This works quite nicely and is organized in a way such as that the saved resources are organized on the page by type under a corresponding header. I can bring this to Nick and Michael's attention for the ok if necessary.

1. go to `products/12/edit/multimedia` and verify you can add and submit for each attachment type
2. verify each addition shows up under the respective attachment type header
3. verify persisted attachments can be deleted
4. verify image attachments can be cropped and the changes to the image persist

Update: since the some logic involved in the practice editor workflow has been updated it's important to go through the workflow and check that everything still functions as expected:
1. For an existing practice, go through the editor and verify updating, adding, and removing data and attachments works as expected
2. Pay close attention to: categories, awards, and resources.

Screenshot 2024-09-23 at 4 52 30 PM

## Acceptance criteria
- [ ]

## Definition of done


app/controllers/products_controller.rb
```diff
@@ -55,8 +66,14 @@ def product_params
...
-      va_employees_attributes: [:id, :name, :role, :_destroy]
-      )
+      va_employees_attributes: [:id, :name, :role, :_destroy],
+    )
+  end
+
+  def multimedia_params
+    params.require(:practice).permit(
+      practice_multimedia_attributes: permitted_dynamic_keys(params[:practice][:practice_multimedia_attributes])
+    )
```
PhilipDeFraties, line 76: I left the multimedia params as being nested under the `:practice` key, this is because the nested forms that are shared directly with the practice editor are a bit tough to untangle and making them dynamic depending on parent type will take extra time (although I'm sure it is doable). I'm on the fence about how egregious this is..

app/assets/javascripts/practice_editor_overview.es6
camillevilla, line 1: I'm wondering if we should extract this to an es6 file that's unique to products. e.g. Products will not have problems, solutions, results so we won't need some of these listeners

camillevilla: 👍 after discussion with @PhilipDeFraties, I think that just renaming this file to make it clearer is uses shared utilities is fine, e.g. `practice_multimedia_utils.es6`

camillevilla (approved):

camillevilla (changes requested):
Hmm, looks like the specs for the Implementation section are not happy with the change.

PR #1046, opened 2024-09-25
Unpin mechanize gem & move to development group

camillevilla:
### JIRA issue link
N/A, related to #1043 

- Moves a gem that we only use for development workflows into the development group


nuel247 (approved):
Looks good to me

PhilipDeFraties (approved):

PR #1051, opened 2024-09-28
DM-5220: Add siderail with Tags to Product page

camillevilla:
### JIRA issue link
[DM-5220](https://agile6.atlassian.net/browse/DM-5220)

 - Adds a siderail to the Product show page
 - Create a class for the customized USWDS `Tags` used on the show pages to make it clearer these are modded components
 - Utilize existing practice page JS for see more / see less logic and showing the Tags section in different parts of the page. (Note: I've left this in the same file for now. I'll decide if all the functions can be in a shared partial after I've finished [DM-5069 for updating the in page nav](https://agile6.atlassian.net/browse/DM-5081))
 - Move `Partners` further up the page in response to client feedback / updated show page designs

1. Go to `/admin/products/` and choose a product to edit
2. Go to the product's description page and add 11 tags to it.
3. Go to the product's show page and confirm that a `See more` link is rendered and works as expected
4. Resize the page to tablet width. Confirm the tags disappear from the siderail (but the whitespace is maintained - this is where in-page nav will go later). The tags should now appear at the bottom of the show page.
5. Resize the page to mobile width. Confirm the tags are still showing at the bottom of the page. The siderail should no longer be there (e.g. the main product content should take up the whole width).

Screenshot 2024-09-27 at 6 51 56 PM

### DIFF
--- a/spec/factories/category_practices.rb
+++ b/spec/factories/category_practices.rb
@@ -2,8 +2,12 @@ FactoryBot.define do
   factory :category_practice do
     association :category
 
-    factory :category_practice_for_product do
+    trait :for_product do
       association :innovable, factory: :product
+    end
+
+    trait :for_practice do
+      association :innovable, factory: :practice
     end
   end
 end
--- a/spec/factories/category_practices.rb
+++ b/spec/factories/category_practices.rb
@@ -1,6 +1,9 @@
 FactoryBot.define do
   factory :category_practice do
     association :category
-    association :innovable, factory: :practice
+
+    factory :category_practice_for_product do
+      association :innovable, factory: :product
+    end
   end
 end
--- a/spec/factories/products.rb
+++ b/spec/factories/products.rb
@@ -67,5 +67,17 @@ FactoryBot.define do
         end
       end
     end
+
+    trait :with_tags do
+        sequence(:name) { |n| "Sample Product with Tags #{n}" }
+        after(:create) do |product|
+        create_list(:category, 11).each do |category|
+          CategoryPractice.create(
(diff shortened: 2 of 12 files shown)


spec/factories/products.rb
```diff
@@ -66,5 +66,17 @@
...
+
+    trait :with_tags do
+        sequence(:name) { |n| "Sample Product with Tags #{n}" }
+        after(:create) do |product|
+        create_list(:category, 11).each do |category|
+          CategoryPractice.create(
+            innovable: product,
+            category: category
+          )
+        end
```
PhilipDeFraties, line 78: with the suggested change to the `CategoryPractice` factory above, this can be updated to:

```
    create_list(:category, 11).each do |category|
      create(:category_practice, :for_product, innovable: product, category: category)
    end
```

PhilipDeFraties (changes requested):
looks good and works as described, just one small change request I left in the factories

spec/factories/category_practices.rb
```diff
@@ -1,6 +1,9 @@
 FactoryBot.define do
   factory :category_practice do
     association :category
-    association :innovable, factory: :practice
+
+    factory :category_practice_for_product do
+      association :innovable, factory: :product
+    end
```
PhilipDeFraties, line 7: to keep things copasetic with the factories I think this would be better off as a trait:

```
trait :for_product do
	association :innovable, factory: :product
end

trait :for_practice do
	association :innovable, factory: :practice
end
```

PhilipDeFraties (approved):

PR #1052, opened 2024-09-30
DM-5231: Add Vendor links to bulk import file

camillevilla:
### JIRA issue link
DM-5231: Add Vendor links to the bulk import file

## Acceptance criteria
- [ ]

## Definition of done


app/views/products/show.html.erb
```diff
@@ -29,7 +29,7 @@
 		[:description, "Executive Summary"],
 		[:main_display_image, "Main Product Image", {heading: false}], # replace this with partial to render
 		[:item_number, "Item Number"],
-		[:vendor, "Vendor"],
+		[:vendor, "Vendor", {content: @product.vendor_link.present? ? link_to(@product.vendor, @product.vendor_link, class: ".usa-link .usa-link--external") : @product.vendor }],
 		[:duns, "DUNS"],
 		[:practice_partners, "Partners", {content: @product&.practice_partners.pluck(:name).join(', ')}],
```
PhilipDeFraties, line 34: Looks like I missed this in the initial implementation of this partial, but the db query in the view here (as well as calling `send` in the block below) is not aligned with best practices.. 

In general the logic in this `section` tag would be best moved into the controller and broken up into helper methods. Functionally this works but I would really like to circle back and refactor this a bit before merging the feature branch into master.

PhilipDeFraties (approved):

app/views/products/show.html.erb
camillevilla: Yeah, that's fair. Now that the data mapping has mostly settled down, we can move this data structure and these DB queries over to a helper soon.

PR #1084, opened 2024-10-21
IPM feature updates after initial QA

camillevilla:
- Removes lorem ipsum placeholder description and provides URL guidance
- Fix typo in Wareologie Parallel Bars video description
- Address bullet gem warning in `/admin/products`
- Add more associations to model specs for models that were changed to polymorphic
- Remove migrated CSS that was clipping product titles
- Tweak whitespace in editor
- Update copy in editor 
- Change mentions of 'Wavier' to Form
- Add an "Edit product" button to the show page


app/views/practices/introduction_forms/_image_editor.html.erb
```diff
@@ -13,7 +13,13 @@
...
   <div class="dm-cropper-boundary grid-col-12 margin-top-2" data-model-type="<%= practice.class.name.downcase %>">
     <label class="dm-cropper-upload-image-label line-height-26" for="practice_main_display_image">
-      Choose a photograph to represent the Innovation in Search Results.
+      <%=
+        if practice.is_a?(Practice)
+          "Choose a photograph to represent the Innovation in Search Results."
+        else
+          "Choose a photograph to represent the Product"
+        end
+      %>
```
PhilipDeFraties, line 22: @camillevilla what do you think of this change for the `main_display_image` input label text as per Michael's feedback? I figured since the products are not yet included in search results this would make sense for now

camillevilla: 👎 I would keep the text as is so we don't have to remember to change this field during Search Results update.

app/views/products/show.html.erb
```diff
@@ -37,12 +37,15 @@
 			<% if @search_terms.any? %>
 		        <%= render partial: 'practices/show/desktop_partials/search_terms', locals: { search_terms: @search_terms } %>
 		  <% end %>
+		  <% if current_user.present? && (current_user&.has_role?(:admin) || current_user == @product.owner || is_user_an_editor_for_innovation(@product, current_user))  %>
+		  	<%= link_to '<i class="fas fa-edit margin-right-1"></i>'.html_safe + 'Edit product', product_editors_path(@product), class: 'usa-button usa-button--secondary margin-top-4 width-full', id: '#show-product-edit-btn' %>
+		  <% end %>
```
PhilipDeFraties, line 42: I think it would be cleaner and more conventional to move this conditional into the controller to be utilized as an instance variable:

```
class ProductsController < ApplicationController
  def show
    @search_terms = @product.categories.get_category_names
    @can_edit_product = current_user.present? && (current_user.has_role?(:admin) || 	 			 	 	 
 	current_user == @product.owner || is_user_an_editor_for_innovation(@product, 
      	current_user))
    render 'products/show'
  end
end
```

PhilipDeFraties (approved):
lgtm!

PR #1147, opened 2024-12-20
Feature: Rails 7

PhilipDeFraties:
Updates `rails` gem to 7.0

Numerous updates to various config files via app:update command.

I went through and reviewed each file updated line by line, reverting changes that needed to be kept (ie keeping the settings for redis caching in the environment config files). Many changes were just adding/updating commented instructions and examples in the config files or replacing single quotes with double.

One notable change was the adding of default precision values for timestamp columns on in the schema, which occurred automatically in the process of running the `app:update` command and unfortunately removed the existing precision settings that needed to be subsequently added back via migration in the second commit. Note: the va_facilities and visns tables' timestamp columns were previously set to a precision of 11 which is (and was) outside the supported range of 0-6 so I reimplemented at the max of 6. Going forward on this version of rails, any time there is an update to the schema there it will automatically add `precision: nil` to any timestamp column that doesn't already have it explicitly set. There is no specific place in the documentation that explains this change, the best I could find is [this issue where the change is discussed](https://github.com/rails/rails/issues/44571).

Another change of note is the new `cookie_rotator` initializer added to ensure backwards compatibility with cookies generated prior to the upgrade which includes a change to the key generator from SHA1 to SHA256. The initializer implemented was taken directly from the rails upgrade documentation seen [here](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#key-generator-digest-class-change-requires-a-cookie-rotator).

## Acceptance criteria
- [ ]

## Definition of done


config/environments/production.rb
```diff
@@ -60,9 +60,9 @@
   # Use a different cache store in production.
   config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL", "redis://localhost:6379/0") }
 
-  # Use a real queuing backend for Active Job (and separate queues per environment)
-  config.active_job.queue_adapter     = :sidekiq
-  # config.active_job.queue_name_prefix = "diffusion-marketplace_#{Rails.env}"
+  # Use a real queuing backend for Active Job (and separate queues per environment).
+  # config.active_job.queue_adapter     = :resque
+  # config.active_job.queue_name_prefix = "diffusion_marketplace_production"
```
camillevilla, line 65: @PhilipDeFraties Can you reenable sidekiq and test in local?

PhilipDeFraties: @camillevilla thanks for catching that, added it back and verified it works as expected when using the "Send Email to Innovation Editors" feature

camillevilla (approved):

